PR #118, opened 2013-03-25
Falsey defaults

justinabrahms:
Allow false-y defaults for fields.
    
A good example of this is a BooleanField which has a default of
False. Without the patch it comes back as None.

### DIFF
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -54,7 +54,7 @@ class BaseType(object):
     
         value = instance._data.get(self.field_name)
 
-        if value is None and self.default:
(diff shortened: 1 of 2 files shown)


bintoro:
As of 1c69ae8f28e70e923ce5413ac247e405565634b8, the implicit default value is `Undefined`.

What this means is that `None` is now also a legal default, and it's possible to make a distinction between NULL and *nothing*. The standard behavior remains such that all fields are initialized to `None` upon instance creation, but setting `init_values=False` will prevent this. Then, only those fields that explicitly declare `default=None` will be initialized to `None`.

PR #124, opened 2013-05-22
Split input data from valid data

lkraider:
Separate new input data from already validated data. This speeds up validation
since it only needs to be run on new data. A couple of data integrity tests are
added too.

### DIFF
--- a/schematics/models.py
+++ b/schematics/models.py
@@ -202,7 +202,14 @@ class Model(object):
             Filter output by a specific role
 
         """
-        return serialize(self, role)
+        try:
+            raw_data = self._raw_data
+            self.validate()
+        except ModelValidationError:
+            pass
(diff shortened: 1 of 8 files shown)


bintoro:
Now that #374 has been merged, I think it's time to finally think about this and reimplement it on top of the current version.

It seems really straightforward, actually. Now that we have `Model.import_data`, `Model.validate` should just delegate to it if it receives additional data. I have already reimplemented `import_data` locally to support recursive updates.

I have also written a tentative patch to stop storing `Undefined` objects in `_data`. Missing keys are cleaner in some respects, and membership tests are really fast. I've been on the fence about it, but splitting `_data` into valid and untrusted portions would seem to tip the scales in favor.

lkraider:
Superceded by #466

PR #129, opened 2013-06-05
Append model fields

lkraider:
This adds a method to append new fields to a model class.
Additionally, adds a copy method so that type instances can be copied.

Example:

``` python
class Person(Model):
    name = StringType()

Person.append_field('gender', Person.name.copy())
```

### DIFF
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -116,6 +116,12 @@ class BaseType(object):
     def __call__(self, value):
         return self.to_native(value)
 
+    def copy(self):
+        copy = object.__new__(self.__class__)
+        copy.__dict__.update(self.__dict__)
+        copy.validators = self.validators[:]
+        return copy
+
     @property
     def default(self):
         default = self._default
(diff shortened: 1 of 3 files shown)


mjrk:
+1
see https://github.com/schematics/schematics/issues/330

schematics/types/base.py
```diff
@@ -116,6 +116,12 @@ def __init__(self, required=False, default=None, serialized_name=None,
     def __call__(self, value):
         return self.to_native(value)
 
+    def copy(self):
+        copy = object.__new__(self.__class__)
+        copy.__dict__.update(self.__dict__)
+        copy.validators = self.validators[:]
```
bintoro: Shouldn't `choices` and `messages` be copied as well?

lkraider:
It's been a while since this was made, it needs updating to the latest schematics version. I am still using the 0.9.x version in production so I didn't have a need to change it yet.

lkraider:
This is much different with Schema now, so the OP will have to create a new PR if he still wants this functionality ;)

PR #134, opened 2013-06-18
Pass old value to field validation

lkraider:
This allows state machine validation, that only accept a value if it
corresponds to some valid state transition.

This could previously be implemented by model-level validators, but it makes
sense to add it as a feature of the base types, since you don't need full
instance-access for that, seeing that the validate function already accepts a
context dictionary of previous valid data.

### DIFF
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -134,7 +134,7 @@ class BaseType(object):
         """
         return value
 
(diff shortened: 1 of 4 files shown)


lkraider:
This functionality is mostly available through the serializable setters now.

Issue #167, opened 2013-09-26
Make an `AbstractType`


j2labs:
Make an AbstractType which has no validation and allows any kind of data to be stored.  This makes the choice to not use a specific type explicit and allows people to store freeform data.

If the data is freeform, I believe this type should allow the `to_native` and `to_primitive` functions to be defined at instance creation time.  This means each instance of an AbstractType can be configured for different behavior.

WyseNynja:
Any suggestions for how to do this? I have a sqlalchemy.JSON field that is a nested dict with a somewhat complicated structure. I don't care about validating the data besides being sure its a valid dict. Is a custom type the right thing to do?

Jbat1Jumper:
Same here. 
In OpenAPI I'm having an object field with _additional properties_ defined in the next way: 
```yaml
data:
    type: object
    additionalProperties: true
```
So it matches agains any JSON object but I can't find a way to define this in schematics.

lkraider:
@Jbat1Jumper there is no strict JSON type, but you can do type that accepts anything by using BaseType()

@WyseNynja If you want to make sure it is a dictionary, you can create a DictType(BaseType)

adihat:
> @Jbat1Jumper there is no strict JSON type, but you can do a field that accepts anything by using BaseType()
> 
> @WyseNynja If you want to make sure it is a dictionary, you can create a DictType(BaseType)

Is there a way to have a JSON with only 2-3 supported types?
Currently `DictType(BaseType)` supports any type. I wanted to restrict this to lets say, NumberType and StringType.
So basically I am looking at something like a `Union` of types as follows:
```python
class MyModel(Model)
    field1 = DictType(UnionType(StringType, NumberType))
```

is this possible or is there a workaround?

PR #185, opened 2013-12-01
AbstractType, AnyPrimitiveType, & AnySimplePrimitiveType

estebistec:
For issue #167
- Define AbstractType with some basic conversion to primitive form
- Include hooks for converting non-primitive teams via simple methods
- Define AnyType(AbstractType), that validates values of any primitive type
- Define SimpleValueType(AbstractType), that validates values of any simple (non-compound) primitive type


tatiana:
@j2labs and @kracekumar: couldn't find AbstractType in schematics==1.0.3 nor in develop branch. Is there some new way of having this behavior?

PR #193, opened 2013-12-12
Primitive types

estebistec:
Add primitive types accepting values of any simple or compound primitive JSON type.

### DIFF
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -223,8 +223,8 @@ class BasePrimitiveType(BaseType):
         if isinstance(value, self.simple_primitive_types):
             return value
 
-        # Get a type-specific converter, or the fallback convert, and apply it to the value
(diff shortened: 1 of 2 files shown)


tatiana:
@j2labs: I was searching for the AbstractType in schematics, but I couldn't find it nor in 1.0.3, nor in the develop branch. Is there some other way to achieve this behavior?

jmsdnns:
I am not sure. I haven't maintained Schematics for a little while. Perhaps @kracekumar can advise?

jmsdnns:
I used to go by @j2labs but I go by @jmsdnns instead now. I also do'nt hack Python anymore and only work on Objective-C or Swift. :(

tatiana:
No problems, @jmsdnns :) Thanks a lot for the quick response!

BTW: I'm using `BaseType` and apparently it works in the expected way.

PR #205, opened 2014-01-28
Changed validaiton loop to assure validation order

### LINKED ISSUE #204: Fields order is not preserved during validation
The docs state: "The order whith which fields are declared is preserved inside the model." [Link](https://schematics.readthedocs.org/en/latest/usage/validation.html?highlight=order%20preserved#model-level-validation). However, it's not actually true... 

Model stores its fields as an OrderedDict, so the order is fixed / preserved. However, validation process is a loop over [_data_ to be validated](https://github.com/j2labs/schematics/blob/master/schematics/validate.py#L76), not fields set. Data comes to validate() function from [import_loop()](https://github.com/j2labs/schematics/blob/master/schematics/validate.py#L41) that for its turn considers data as a regular [dict](https://github.com/j2labs/schematics/blob/master/schematics/transforms.py#L52)

Example below shows that validate_call_me() method gets hit _before_ validate_name() although the fields are defined in reversed order.

```
from schematics.types import StringType, BooleanType
from schematics.exceptions import ValidationError

class Signup(Model):
    name = StringType()
    call_me = BooleanType(default=False)
    
    def validate_name(self, data, value):
        if data['name'] == u'Brad':
            return u'Joe'
        return value
    
    def validate_call_me(self, data, value):
        print 'validate_call_me: {}'.format(data['name'])
        if data['name'] == u'Joe' and data['call_me'] is True:
            raise ValidationError(u'He prefers email.')
        return value
## -- End pasted text --

In [32]: Signup({'name': u'Brad', 'call_me': True}).validate()
validate_call_me: Brad
```

Please, let me know if it makes sense.

Gennady-Andreyev:
Fixes #204 
cls._fields is an instance of OrderedDict while data is a regular dictionary

### DIFF
--- a/schematics/validate.py
+++ b/schematics/validate.py
@@ -73,8 +73,9 @@ def _validate_model(cls, data):
         Errors of the fields that did not pass validation.
     """
     errors = {}
-    for field_name, value in data.items():
-        if field_name in cls._validator_functions:
+    for field_name, field in cls._fields.iteritems():
+        if field_name in cls._validator_functions and field_name in data:
+            value = data[field_name]
             try:
                 context = data
                 if hasattr(cls, '_data'):
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -171,6 +171,30 @@ def test_multi_key_validation_part_two():
         Signup({'name': u'Brad', 'call_me': True}).validate()
 
 
+def test_multi_key_validation_fields_order():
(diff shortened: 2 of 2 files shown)


kracekumar:
Can you rebase and update the pull request ? I will merge this.

csojinb-picwell:
This can be closed now, because the code was replicated in #308

PR #207, opened 2014-01-28
Implemented update of cleaned data according to validator's return value

### LINKED ISSUE #206: Whatever returned from model validation methods gets disregarded
According to docs / examples, model-level validation methods return ["clean" values](http://schematics.readthedocs.org/en/latest/usage/validation.html#model-level-validation). But current implementation of _validate_model() [doesn't expect any return value](https://github.com/j2labs/schematics/blob/master/schematics/validate.py#L82), so whatever returned from model validation methods gets disregarded.

Gennady-Andreyev:
Fixes #206


bintoro:
Superseded by #355.

PR #230, opened 2014-05-14
Message normalization

kstrauser:
All messages:

- Are wrapped in u""
- End with a period
- Use '{n}'.format(value) string formatting


jokull:
@kstrauser can you merge with master and squash into one commit?

kstrauser:
@jokull Done!

PR #236, opened 2014-06-25
Set `None` on a field that's a compound type should honour that semantics

kevinjqiu:
When a compound field is set to None, set the value
of that field to `None` instead of creating an empty `ModelType` of that
compound object.

### DIFF
--- a/schematics/models.py
+++ b/schematics/models.py
@@ -46,7 +46,10 @@ class FieldDescriptor(object):
         Checks the field name against a model and sets the value.
         """
         field = instance._fields[self.name]
-        if not isinstance(value, Model) and isinstance(field, ModelType):
+        if all((
+                value is not None,
+                not isinstance(value, Model),
+                isinstance(field, ModelType))):
             value = field.model_class(value)
         instance._data[self.name] = value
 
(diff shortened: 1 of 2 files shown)


jhorman:
It seems like there is still no movement on this, or on the Undefined proposal. My company is running into this issue. Not being able to unset a ModelType field is a fairly large problem. We have to work around it by setting `_data[field] = None` directly. cc @j2labs

bintoro:
As I commented here a year ago, I wholeheartedly agree with this PR.

In fact, I've always wondered why `FieldDescriptor.__set__()` coerces `ModelType` values and nothing else.

```python
if not isinstance(value, Model) and isinstance(field, ModelType):
  value = field.model_class(value)
```

Why does the code do this?

At the very least `value = None` should be honored, but I'm thinking maybe we can remove the magical model instantiation altogether. Could one of the old-timers please chime in about the reasoning behind this?

ikonst:
Is this going to be merged in? Just ran into this issue too. As @bintoro says, the current behavior is thoroughly baffling.

bintoro:
Yeah let's get this done.

PR #243, opened 2014-08-21
DBRefType for MongoDb contrib

wsantos:
I have been working in a AsyncORM using motor and schematics called
Motorm, and this type would help a lot.


lkraider:
Although I think this could be useful, I don't see this getting updated/supported in the core lib seeing as how little the current contrib is touched.

If someone is interested in maintaining mongodb mappings, a separate project (maybe even under the schematics name?) for a mongo lib would be a better home I think.

PR #267, opened 2014-12-18
Type choices should be list or tuple

kaiix:
It's better to check `choices` type, because string is also a valid option.


schematics/types/base.py
```diff
@@ -180,6 +180,8 @@ def __init__(self, required=False, default=None, serialized_name=None,
         self.required = required
         self._default = default
         self.serialized_name = serialized_name
+        if choices and not isinstance(choices, (list, tuple)):
+            raise TypeError('"choices" must be a list or tuple')
```
hellp: what about sets and frozensets?

kracekumar: Good catch! Open a PR ?

PR #269, opened 2014-12-21
Marshallers and extended serializables for better data manipulation at import/export time

bintoro:
The `serializable` decorator currently provides a simple way for a model to expose a virtual field. However, the mechanism doesn't handle two-way communication.

Consider a model with fields `first_name` and `last_name` that needs to talk to a client that only sends and receives a single value, `full_name`:

``` python
class Person(Model):

    first_name = StringType()
    last_name = StringType()

    @serializable
    def full_name(self):
        return '%s %s' % (self.first_name, self.last_name)
```

The above works for outbound data, but what about the reverse? How to split the incoming value into two concrete fields?

This PR introduces two facilities that provide pre- and post-processing hooks for data manipulation.
### Extended serializables

Instead of decorating a function, decorate a class that encloses functions for both exporting and importing:

``` python
    @serializable
    class full_name:

        def serialize(self):
            return '%s %s' % (self.first_name, self.last_name)

        def deserialize(value):
            first_name, last_name = value.split()
            return {
                'first_name': first_name, 
                'last_name': last_name
            }
```

The wrapper class (`full_name`) is nothing but a closure. It will never be instantiated, and inside the `serialize` function `self` still refers to the `Person` instance. In other words, `serialize()` here is identical to the classic one-way serializable.

The dictionary returned by `deserialize()` will be added to the dataset, and the original incoming field (`full_name`) will be automatically discarded.
### Marshalling/unmarshalling functions

These are functions that can perform arbitrary operations on the entire dataset.

``` python
class Person(Model):

    first_name = StringType()
    last_name = StringType()

    def marshal_full_name(cls, data):
        data['full_name'] = '%s %s' % (data['first_name'], data['last_name'])

    def unmarshal_full_name(cls, data):
        if 'full_name' not in data:
            raise ModelConversionError("Field 'full_name' missing") 
        first_name, last_name = data['full_name'].split()
        new_data = {
            'first_name': first_name, 
            'last_name': last_name
        }
        return (new_data, ('full_name',))
```

An **unmarshaller** (processes inbound data) must return a tuple with two items:
- a dictionary that will be added to the dataset
- an iterable of keys that should be discarded (nothing is dropped automatically).

A **marshaller** (processes outbound data) may do the same, or it can modify the dataset directly and return nothing, as shown above.

Apart from the `(un)marshal_` prefix, the names of the functions are of no consequence (except for error info), since they do not necessarily relate to just a single field. A model-wide pre/post-processor can be named simply `unmarshal` or `marshal`.
### Comparison
- Serializables deal with native values: they are invoked after `to_native()` upon import and before `to_primitive()` upon export.
- Marshalling functions deal with primitive values: they are called before `to_native()` upon import and after `to_primitive()` upon export.
- Serializables always map to **one** exposed field. If you wanted to do the opposite of the example and map multiple incoming values to a single concrete field, a serializable will not do.
- Marshalling is more powerful, since the entire dataset can be freely manipulated.
- The unmarshallers are run before anything else is done to the incoming data. For example, multiple values may be coalesced into one, and the surplus transport fields can be dropped before rogue fields are checked for.
- A serializable is always available as `instance.some_serializable`, whereas marshallers are only evaluated at export time.
### Also changed

Serializables now accept the parameters `default`, `required`, and `deserialize_from`. Note that `required` on a serializable is only enforced on import — not during validation.

The default value for `serialize_when_none` on serializables now matches that on types (changed from `True` to `None`).

### DIFF
--- a/schematics/models.py
+++ b/schematics/models.py
@@ -128,6 +128,12 @@ class ModelMeta(type):
         fields = OrderedDictWithSort()
         serializables = {}
         validator_functions = {}  # Model level
+        marshallers = OrderedDictWithSort()
+        unmarshallers = OrderedDictWithSort()
+
+        main_marshaller = attrs.get('marshal')
+        if callable(main_marshaller):
+            marshallers[name] = main_marshaller
 
         # Accumulate metas info from parent classes
         for base in reversed(bases):
@@ -137,15 +143,27 @@ class ModelMeta(type):
                 serializables.update(metacopy(base._serializables))
             if hasattr(base, '_validator_functions'):
                 validator_functions.update(base._validator_functions)
+            if hasattr(base, '_marshallers'):
+                marshallers.update(base._marshallers)
+            if hasattr(base, '_unmarshallers'):
+                unmarshallers.update(base._unmarshallers)
 
         # Parse this class's attributes into meta structures
         for key, value in iteritems(attrs):
             if key.startswith('validate_') and callable(value):
                 validator_functions[key[9:]] = value
+            if key.startswith('marshal_') and callable(value):
+                marshallers[key[8:]] = value
+            if key.startswith('unmarshal_') and callable(value):
+                unmarshallers[key[10:]] = value
             if isinstance(value, BaseType):
                 fields[key] = value
             if isinstance(value, Serializable):
                 serializables[key] = value
+
+        main_unmarshaller = attrs.get('unmarshal')
+        if callable(main_unmarshaller):
+            unmarshallers[name] = main_unmarshaller
 
         # Parse meta options
         options = mcs._read_options(name, bases, attrs)
@@ -159,6 +177,8 @@ class ModelMeta(type):
         attrs['_fields'] = fields
         attrs['_serializables'] = serializables
         attrs['_validator_functions'] = validator_functions
+        attrs['_marshallers'] = marshallers
+        attrs['_unmarshallers'] = unmarshallers
         attrs['_options'] = options
 
         klass = type.__new__(mcs, name, bases, attrs)
@@ -239,7 +259,7 @@ class Model(object):
         if raw_data is None:
             raw_data = {}
         self._initial = raw_data
-        self._data = self.convert(raw_data, strict=strict, mapping=deserialize_mapping)
+        self._data = self.convert(raw_data, strict=strict, mapping=deserialize_mapping, importing=True)
 
     def validate(self, partial=False, strict=False):
         """
@@ -269,6 +289,7 @@ class Model(object):
         :param raw_data:
             The data to be imported.
         """
+        kw.setdefault('importing', True)
         data = self.convert(raw_data, **kw)
         #[x * 2 if x % 2 == 0 else x for x in a_list]
         del_keys = [ k for k in data.keys() if data[k] is None]
--- a/schematics/transforms.py
+++ b/schematics/transforms.py
@@ -6,6 +6,7 @@ import itertools
 from six import iteritems
 
 from .exceptions import ConversionError, ModelConversionError, ValidationError
+from .types.serializable import Serializable
 
 try:
     basestring #PY2
@@ -30,7 +31,7 @@ except:
 ###
 
 def import_loop(cls, instance_or_dict, field_converter, context=None,
-                partial=False, strict=False, mapping=None):
+                partial=False, strict=False, mapping=None, importing=False):
     """
     The import loop is designed to take untrusted data and convert it into the
     native types, as described in ``cls``.  It does this by calling
@@ -63,6 +64,19 @@ def import_loop(cls, instance_or_dict, field_converter, context=None,
     data = dict(context) if context is not None else {}
     errors = {}
 
+    # Unmarshal
+    if importing:
+        for name, unmarshaller in cls._unmarshallers.items():
+            try:
+                data_, del_ = unmarshaller(cls, instance_or_dict)
+            except ConversionError as exc:
+                errors["Unmarshaller for '%s'" % name] = exc.messages
+            else:
+                for k, v in data_.items():
+                    instance_or_dict[k] = v
+                for k in del_:
+                    del instance_or_dict[k]
+
     # Determine all acceptable field input names
     all_fields = set(cls._fields) ^ set(cls._serializables)
     for field_name, field, in iteritems(cls._fields):
@@ -79,7 +93,13 @@ def import_loop(cls, instance_or_dict, field_converter, context=None,
         for field in rogue_fields:
             errors[field] = 'Rogue field'
 
-    for field_name, field in iteritems(cls._fields):
+    if importing:
+        fields = itertools.chain(iteritems(cls._fields),
+                                 iteritems(cls._serializables))
+    else:
+        fields = iteritems(cls._fields)
+
+    for field_name, field in fields:
         serialized_field_name = field.serialized_name or field_name
 
         trial_keys = _list_or_string(field.deserialize_from)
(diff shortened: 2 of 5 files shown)


UberFarmer:
+1 I very much want marshalling / unmarshalling functionality. Are there any clean(ish) workarounds until this PR is accepted?

In particular, I want to be able to execute custom code when a field is set / retrieved. In other words, exactly the functionality I would get from using properties if it was possible to use properties.

breerly:
+1 - this looks great :)

bintoro:
Thanks for the comments, guys. Now that I've gotten some feedback, I might merge this soonish.

Something I'd like to resolve is whether and how applications should be able to specify the ordering of several marshallers. Of course there are many ways to accomplish this, or we could simply say developers shouldn't write marshallers so that they depend on one anothers' outputs. Anyway, this got me thinking: since the names of the methods have no real significance, is the automatic identification of all "marshal_" methods even that useful — as opposed to just a single master marshalling method? Having only one method to execute (which may or may not be split into several subroutines, of course) would take the ordering question out of the equation.

@UberFarmer — I don't think there's anything to stop you from using regular Python properties and the *serialized_name* parameter on the field to accomplish what the SA code is doing. Except, of course, that the accessor methods are not applied during import or export, which may be what you're asking for here. And I agree this is something to be addressed, since the (un)marshallers are supposed to affect import/export *only*. If there's a transformation that should be *always* applied (on import/export as well as when application code accesses the attribute), then I think the code should be placed in some Schematics-specific accessor just like SA's *hybrid_property*. This way all three cases would be covered:
*    to transform on attribute access only, use regular property
*    to transform on import/export only, use marshaller
*    to transform always, use magic property.

Am I making sense here?

bintoro:
After reflecting on this, the obvious candidate to act as a property-like accessor is the `serializable`. It already handles getting, so all we need to do is modify the "extended serializable" plan a bit:


```python
    @serializable
    class full_name:

        def serialize(self):
            return '%s %s' % (self.first_name, self.last_name)

        def deserialize(self, value):
            self.first_name, self.last_name = value.split()
```

So, instead of returning a dictionary, the deserializer method simply sets the attributes directly on the instance. This is actually a much cleaner solution, and now the setter can be invoked at any time. The import machinery just needs to create a dummy instance on which the setter operates, since the final object doesn't yet exist at that point, but other than that I don't foresee any complexities here.

By the way, I should mention that there's no technical reason (I think...) why the two functions should be enclosed within a class construct. It's also possible to opt for a more canonical syntax:

```python

    @serializable
    def full_name(self):
        return '%s %s' % (self.first_name, self.last_name)

    @full_name.setter
    def full_name(self, value):
        self.first_name, self.last_name = value.split()
```

There are endless ways to name the second decorator; the only requirement is that we need a reference to the main serializable, either as the decorator object or its parameter. For example:

```python
    serializable.set(full_name)
    def full_name(self, value):
        ...
```

UberFarmer:
Personally, I'm not sure there's a compelling reason to support offering different behaviors on import and when setting attributes directly. If this code:
```python
person = Person({'first_name': 'John'})
```
and this code:
```python
person = Person()
person.first_name = 'John'
```
can cause different results, that seems like a huge, huge opportunity for user error. The majority of users  *will* expect these two snippets to behave exactly the same, and thus will use these two forms interchangeably. The confusion and subtle bugs surrounding this will by far outweigh the benefits of the additional extra functionality.

bintoro:
@UberFarmer, the marshalling filter is only intended for wholesale munging of incoming or outgoing data. Particularly now that the serializable will gain property-like setter functionality, the use cases for the two are completely different, so I don't think there will be too much room for confusion.

That said, you again got me thinking about how to improve this thing. See, the main use case for marshalling transformations is when you have to accommodate data sources/consumers outside of your control. Well, obviously there might be a need for more than one such transformation (one system expects *full_name*, the other expects *first_name* and *last_name*). So instead of having a fixed marshalling logic within a model, it would be better represented as a separate adapter-like object that would be invoked at export time:

```python
instance.to_primitive(adapter=LegacySystemMarshaller)
```
I hope this example also makes it clear why I believe it's useful to be able to perform data manipulation at the very boundary of Schematics. Without it, you would always have to adapt the inner workings of your application according to the specifics of the remote system you are talking to.

UberFarmer:
I like this idea. Multiple schemas/adapters for importing/exporting your data seems awesome.

Okay, so here's an example class that includes most of the features we discussed. Let's see if this agrees with you. Syntax can, of course, be changed.

```python
class Person(Model):

    _first_name = StringType(serialize=False)  # Option to not import/export this field, but still validate it

    @serializable
    def first_name(self):
        return self._first_name

    @first_name.setter
    def first_name(self, value):
        if not self.first_initial and len(value) >= 1 :
            self.first_initial = value[0]
        self._first_name = value

    last_name = StringType()
    first_initial = StringType()

    class LegacyAdapter:  
        # Adapter to munge 'full_name' into 'first_name' and 'last_name' on import,
        # and from 'first_name' and 'last_name' into 'full_name' on export.
        # Not sure what the syntax should be here...
```

##### Importing Steps:
* Data & an optional adapter are passed in
* If the adapter is specified, the adapter munges the data
* Construct: For every data field, set that attribute on the model by calling its set method

*Note: This means when no adapter is specified, constructing with data has exactly the same effect as constructing a default model and setting the attributes one by one, as I was worried about in my previous comment*

Thus
```python
Person({'first_name': 'John', 'last_name': 'Smith'})
```
,
```python
Person({'full_name': 'John Smith'}, adapter=LegacyAdapter)
```
, and
```python
Person({'first_name': 'John', 'last_name': 'Smith', 'first_initial': 'J'})
``` 
would all create exactly the same model. The first two would both call the `first_name` property setter, thereby populating the `first_initial` field.

##### Exporting Steps:
* Optional adapter is passed in
* Create export data -- for every attribute on the model, get that attribute by calling its get method
* If the adapter is specified, the adapter munges the data
* Data is returned.

Fields returned for regular export: `first_name`, `last_name`, `first_initial`
Fields returned for export with adapter: `full_name`, `first_initial`

##### Attribute Access:
* Serializable getters and setters activate on property access, just as if they were normal Python properties.

alexhayes:
Is there any plan to implement something as @UberFarmer [has suggested](https://github.com/schematics/schematics/pull/269#issuecomment-69684490)?

alexhayes:
A quick sketch which serves my purposes;

```python


class AdapterModel(Model):

    def __init__(self, raw_data=None, adapter: 'BaseAdapter'=None, *args, **kwargs):
        if adapter is not None:
            raw_data = adapter(self).to_raw_data(raw_data)

        super(AdapterModel, self).__init__(raw_data, *args, **kwargs)

    def to_adaptation(self, adapter):
        return adapter(self).from_model()


class BaseAdapter(object):

    def __init__(self, model: AdapterModel):
        self.model = model


class Product(AdapterModel):
    ...
    ...


    class LegacyAdapter(BaseAdapter):

        def to_raw_data(self, obj: dict) -> dict:
            first_name, last_name = obj['full_name'].split(' ')
            return dict(first_name=first_name, last_name=last_name)

        def from_model(self) -> dict:
            return dict(full_name='%s %s' % (self.model.first_name, self.model.last_name))

```

And in use as @UberFarmer as defined above with the following;

```python
product = Product()
...
...
legacy = product.as_adaptation(Product.LegacyAdapter)
```

Thoughts?

lkraider:
I have a setter for serializable fields as @UberFarmer mentioned in my branch here:
https://github.com/schematics/schematics/compare/development...nKey:new-serializable-setter?expand=1
(old) PR here: https://github.com/plain-vanilla-games/schematics/pull/4

bintoro:
Just a heads up, I'm trying to get v2.0 finished as soon as possible. Let's look at this right after. I definitely still want to get this functionality in.

lkraider:
With #466 merged, serializable.setter now enables this - see for example: https://github.com/schematics/schematics/blob/58296cb4cd1f41ad6a63c6ef890c70c1b5b0fb17/tests/test_schema.py#L117

lkraider:
Closing based on current development branch having similar functionality.

bintoro:
What? The scope of this feature suggestion goes far beyond serializable setters.

lkraider:
@bintoro I might have glossed over too fast over the issue, but from the examples cited here they seem to fit within the mutation step provided by instance setters (ie: they apply during import and export).

PR #270, opened 2014-12-25
add a one paramters to Models to_native/to_prim/.. method  to deside whether export the serilized fields.

hackrole:
I think this was some need features. 
while the user validate the data and want save it to db, he didnot may also want to save the serilized fields.
If I have to define the options on every models, I think it is a litter bad.

### DIFF
--- a/schematics/models.py
+++ b/schematics/models.py
@@ -288,10 +288,11 @@ class Model(object):
         """
         return convert(self.__class__, raw_data, **kw)
 
-    def to_native(self, role=None, context=None):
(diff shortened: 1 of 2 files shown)


bintoro:
Could you give an example of why this is needed specifically with regard to serializables? Why define serializables in the first place if you don't want them serialized? You can use a regular Python property instead.

I realize sometimes you might want to exclude certain fields from the output, but that scenario is in no way limited to serializables. It would be better to be able to exclude *any* field (including serializables). And this use case will in fact be addressed by the upcoming marshaller/adapter functionality.

PR #272, opened 2015-01-07
Update 'url' in setup.py

grzn:
Project has moved to a different account on GitHub, but on PyPI the url is still the same


grzn:
This is already implemented on the master branch.
Its confusing that the default branch on GitHub is _development_, while the most recent branch is _master_; please change the default.

Issue #279, opened 2015-02-12
Broken import in models.py


csojinb-picwell:
Since the update to 1.0-1, I'm getting the following error that seems to be internal to the library:
```    
from schematics.types.compound import ListType
../../../virtualenv/python2.7.9/lib/python2.7/site-packages/schematics/types/compound.py:9: in <module>
    from ..models import Model
../../../virtualenv/python2.7.9/lib/python2.7/site-packages/schematics/models.py:435: in <module>
    from .types.compound import ModelType
E   ImportError: cannot import name ModelType
```
Looks like a circular import issue?

kracekumar:
It is fixed in `1.0.2`. 

```
In [2]: schematics.version_info
Out[2]: ('1', '0', '2')

In [3]: from schematics.types.compound import ListType

```

Thanks for reporting.

PR #280, opened 2015-02-12
Issue 279

kracekumar:
- Fix the circular import issue.

### DIFF
--- a/schematics/models.py
+++ b/schematics/models.py
@@ -61,6 +61,7 @@ class FieldDescriptor(object):
         """
         Checks the field name against a model and sets the value.
         """
+        from .types.compound import ModelType
         field = instance._fields[self.name]
         if not isinstance(value, Model) and isinstance(field, ModelType):
             value = field.model_class(value)
@@ -430,6 +431,3 @@ class Model(object):
 
     def __unicode__(self):
         return '%s object' % self.__class__.__name__
-
-
-from .types.compound import ModelType


bintoro:
Sooo this is interesting.

A few days ago, I noticed there was an import statement inside `FieldDescriptor.__set__()`, so I promptly moved it to the end of the file (https://github.com/schematics/schematics/commit/860d71778421981f284c0612aec665ebf0cfcba2). The reason is that calling `import` is expensive even if the module has already been loaded.

Then I found this PR, which is the exact reverse of the change I made.

I wonder why the reporter of #279 had run into a circular import because clearly it had worked for other people. @kracekumar do you have any idea of what went wrong there?

Let's try the bottom import again with v1.1.1. At least all tests are passing. If people start seeing the import issue again, then we'll figure it out, but IMO an import statement cannot live inside one of the most frequently called functions in the library.

kracekumar:
@bintoro I didn't do poor introspection.

Sure. We can move that to bottom of the function. But imports are cached in `sys.modules` and on next run it will be fetched from `sys.modules` correct ? I agree with you to keep imports outside function definitions to avoid any run time issues.

bintoro:
Yeah, imports are cached, but simply invoking the `import` statement nearly tripled the execution time of the setter method.

I just don't understand why that one guy ran into an import issue while the rest of us didn't.

Apparently I was the one who originally created the circular import in https://github.com/schematics/schematics/commit/03b2fd97fb47c00e8d667cc8ea7254cc64d0f0a0 when I needed to use `Model` in `compound.py`. I solved it by moving the import in `models.py` from the top to the bottom of the file, which was fine until for some reason it didn't work for the person who opened #279.

bintoro:
@kracekumar I've got it. The circular import only occurred when `types.compound` was imported before the `models` module. The other way around it has always worked fine. I did the bottom import trick in `types/compound.py` as well, and now it's good both ways. :tada:

PR #281, opened 2015-02-12
Update Model.__eq__

toddsifleet:
* Model.__eq__ now returns NotImplemented if the other object is not the
  same class.  This allows the other objects __eq__ to be checked.  This
  is particularly important for testing (e.g. mock.Any, equals).


toddsifleet:
Without this

UberFarmer:
+1

PR #288, opened 2015-03-12
Add ability to get datetime from timestamp

charlax:
Also fixes some regression related to `ModelType`.

### DIFF
--- a/schematics/models.py
+++ b/schematics/models.py
@@ -165,12 +165,16 @@ class ModelMeta(type):
 
         # Add reference to klass to each field instance
         def set_owner_model(field, klass):
-            field.owner_model = klass
-            if hasattr(field, 'field'):
+            try:
+                field.owner_model = klass
                 set_owner_model(field.field, klass)
+            except AttributeError:
+                pass
+
         for field_name, field in fields.items():
             set_owner_model(field, klass)
-            field.name = field_name
+            if hasattr(field, 'name'):
+                field.name = field_name
 
         # Register class on ancestor models
         klass._subclasses = []
@@ -421,7 +425,7 @@ class Model(object):
         return u"<%s: %s>" % (class_name, obj)
 
     def __str__(self):
-        return '%s object' % self.__class__.__name__        
+        return '%s object' % self.__class__.__name__
 
     def __unicode__(self):
         return '%s object' % self.__class__.__name__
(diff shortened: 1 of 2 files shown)


abeinstein:
+1

schematics/models.py
```diff
@@ -165,12 +165,16 @@ def __new__(mcs, name, bases, attrs):
 
         # Add reference to klass to each field instance
         def set_owner_model(field, klass):
-            field.owner_model = klass
-            if hasattr(field, 'field'):
+            try:
+                field.owner_model = klass
                 set_owner_model(field.field, klass)
+            except AttributeError:
```
kracekumar: Any reason to use `Exception` ?. `pass` is never accepted.

kracekumar:
Since this patch uses `arrow`, it is better to stay out of library. Please create an appropriate issue you're facing and then we can discuss on the solution.

schematics/models.py
charlax: What do you mean? I'm using a specific `AttributeError` exception. I can write a regression test for this if you want.

bintoro:
Hey @charlax, sorry about your PRs being neglected. I totally agree that the datetime types should be much more flexible and powerful. If this issue is still relevant to you, let's first discuss the appropriate strategy in #289. I'll post some thoughts there shortly.

Just to get something out of the way first: this PR includes a change to `ModelMeta`, but I don't understand what it's meant to do. Could you please elaborate? On lines 168–169, the `hasattr` conditional is replaced with a `try – except AttributeError`, which seems to accomplish the exact same thing. On line 176, the added `hasattr` doesn't appear to serve any purpose, as surely you don't need to check for the presence of `field.name` before simply assigning to it. So, what am I missing?

schematics/types/base.py
```diff
@@ -719,6 +730,9 @@ def to_native(self, value, context=None):
             return value
 
         for fmt in self.formats:
+            if fmt == self.TIMESTAMP_FORMAT:
+                return arrow.get(value)
```
bintoro: Let's not add a whole new dependency for such a simple conversion. `datetime.utcfromtimestamp(float(value))` does the same thing. Also, this should look out for exceptions and ignore, in case some other format matches the input.

bintoro:
Superseded by #358.

Issue #289, opened 2015-03-12
Improve flexible handling of datetime representation


charlax:
Schematics does not allow efficient and flexible handling of datetime representation.

**`DateTimeType` does not allow parsing from timestamp**

```python
In [3]: DateTimeType().to_native(1426494600)
---------------------------------------------------------------------------
ConversionError                           Traceback (most recent call last)
<ipython-input-3-da60245683cd> in <module>()
----> 1 DateTimeType().to_native(1426494600)

/opt/Personal/Dropbox/Code/schematics/schematics/types/base.py in to_native(self, value, context)
    727             except (ValueError, TypeError):
    728                 continue
--> 729         raise ConversionError(self.messages['parse'].format(value))
    730
    731     def to_primitive(self, value, context=None):

ConversionError: [u'Could not parse 1426494600. Should be ISO8601.']
```

That's annoying because deserializing should be permissive, especially when multiple input format can be used. This is what #288 aims to fix.

**`TimeStampType` has multiple problems**

It does not override `to_native`, which means that it's basically useless and returns the same error as `DateTimeType` (reported in #271 as well):

```python
In [1]: from schematics.types.temporal import TimeStampType

In [2]: TimeStampType().to_native(1426494600)
---------------------------------------------------------------------------
ConversionError                           Traceback (most recent call last)
<ipython-input-2-72f40cf29e6e> in <module>()
----> 1 TimeStampType().to_native(1426494600)

/opt/Personal/Dropbox/Code/schematics/schematics/types/base.pyc in to_native(self, value, context)
    727             except (ValueError, TypeError):
    728                 continue
--> 729         raise ConversionError(self.messages['parse'].format(value))
    730
    731     def to_primitive(self, value, context=None):

ConversionError: [u'Could not parse 1426494600. Should be ISO8601.']
```

The more important problem is that it ties a type to its deserialization format. We should decouple serialization/deserialization and the type here. 

Let's say I expect a date & time to be deserialized from either timestamp or iso8601 format, and serialized in iso8601. Neither `TimeStampType` nor `DateTimeType` fits the bill here, because:

* `DateTimeType` does not parse timestamps
* `TimeStampType` (assuming we fix it `to_native` method, I have a fix for it as well) only serializes timestamps.

I think the proper fix would be to have `DateTimeType` parse timestamps.

bintoro:
[...]

PR #290, opened 2015-03-12
Override TimeStampType.to_native

### LINKED ISSUE #271: TimeStampType doesn't import
TimeStampType doesn't define to_native, so it calls DateTimeType's to_native, which expects an ISO8601 string instead.

charlax:
Fixes #271. Goes toward #289.


bintoro:
Certainly we want `TimeStampType` to retain the ability to accept other formats than timestamps, no? Fixing the underlying `DateTimeType` seems like the way to go.

PR #295, opened 2015-03-19
Add parameter to pass in a list of Role names for use in export

tim-d-blue:
This provides a simpler method to combine multiple roles on export. Given a simple list of roles, when export is called a list of roles can be provided and the export will combine the roles and filter fields appropriately. Therefore dynamic role combination is available. Blacklist roles will override whitelist roles.

### DIFF
--- a/docs/usage/exporting.rst
+++ b/docs/usage/exporting.rst
@@ -312,6 +312,44 @@ results without having to specify ``role`` in the export function.
       }]
   }
 
+Optionally a list of roles can be provided for export instead. The combination
+of the all the roles will be made and used to filter fields. Blacklist roles
+override whitelist roles in this case.
+
+::
+
+    class User(Model):
+        id = IntType(default=42)
+        name = StringType()
+        email = StringType()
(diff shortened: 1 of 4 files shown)


bintoro:
I like the idea, but couldn't we repurpose the existing `role` parameter instead of introducing a parallel parameter `roles`?

Something like:

```python
if isinstance(role, Role):
    role = (role,)
```

and then the rest of the code could just assume it's an iterable of roles.

Btw, the `__name__` attribute on functions has been around at least since Python 2.5, so I'm thinking this whole thing:

```python
if role and ((hasattr(role.function, "func_name") and
              role.function.func_name == "whitelist")
             or (hasattr(role.function, "__name__") and
                 role.function.__name__ == "whitelist")):
```

can probably be simplified to:

```python
if role and role.function.__name__ == "whitelist"
```

berislavlopac:
Any updates on this PR?

lkraider:
Needs to be refactored according to suggestions before pulling it in.

PR #297, opened 2015-03-20
Support for fields order in serialized format via Options class in model

delicb:
I have implemented simple way of declaring order of fields in serialized output. 
Something like this does not have great technical value, but it is nice to return dictionary always in the same format when using schematics as validation layer for REST API (which is the way that I am using it).

This might not be the best way to implement this, but it reflects just the thing that I need. 

I was not sure what to do with fields that are not specified in `fields_order` list, so I did nothing and in that case order of fields is not guaranteed in any way, just as it is not guaranteed with `dict` (well, fields that are mentioned in `fields_order` will be pushed back, other fields will be left in front, but that is just a side effect).

What do you think?

I am opened to suggestions, if you have any.

### DIFF
--- a/schematics/transforms.py
+++ b/schematics/transforms.py
@@ -6,6 +6,7 @@ import itertools
 from six import iteritems
 
 from .exceptions import ConversionError, ModelConversionError, ValidationError
+from .datastructures import OrderedDict
 
 try:
     basestring #PY2
@@ -157,6 +158,9 @@ def export_loop(cls, instance_or_dict, field_converter,
     else:
         gottago = cls._options.roles.get("default", gottago)
 
+    fields_order = (getattr(cls._options, 'fields_order', None)
+                    if hasattr(cls, '_options') else None)
+
     for field_name, field, value in atoms(cls, instance_or_dict):
         serialized_name = field.serialized_name or field_name
 
@@ -189,9 +193,33 @@ def export_loop(cls, instance_or_dict, field_converter,
 
     # Return data if the list contains anything
     if len(data) > 0:
+        if fields_order:
+            return sort_dict(data, fields_order)
         return data
     elif print_none:
         return data
+
+
+def sort_dict(dct, based_on):
+    """
+    Sorts provided dictionary based on order of keys provided in ``based_on``
+    list.
+
+    Order is not guarantied in case if ``dct`` has keys that are not present
+    in ``based_on``
+
+    :param dct:
+        Dictionary to be sorted.
+    :param based_on:
+        List of keys in order that resulting dictionary should have.
+    :return:
+        OrderedDict with keys in the same order as provided ``based_on``.
+    """
+    return OrderedDict(
+        sorted(
+            dct.items(),
+            key=lambda el: based_on.index(el[0] if el[0] in based_on else -1))
+    )
 
 
 def atoms(cls, instance_or_dict):
--- a/tests/test_ordered_serialization.py
+++ b/tests/test_ordered_serialization.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+from schematics.datastructures import OrderedDict
+from schematics.models import Model
+from schematics.types import StringType, IntType
+
+
+def test_serialize_field_orders_one_two_three():
+    class Example(Model):
+        one = StringType()
+        two = IntType()
+        three = StringType()
+
+        class Options:
+            fields_order = ['one', 'two', 'three']
+
+    p1 = Example({'one': 'a', 'two': 2, 'three': '3'})
+    order = [('one', 'a'), ('two', 2), ('three', '3')]
+    serialized = p1.to_native()
+    assert serialized.items() == order
+
+
+def test_serialize_field_orders_three_two_one():
+    class Example(Model):
+        one = StringType()
+        two = IntType()
+        three = StringType()
+
+        class Options:
+            fields_order = ['three', 'two', 'one']
+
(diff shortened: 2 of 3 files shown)


kracekumar:
I like the idea, there  are quite few useful cases.

schematics/transforms.py
```diff
@@ -156,6 +156,10 @@ def export_loop(cls, instance_or_dict, field_converter,
         raise ValueError(error_msg % (cls.__name__, role))
     else:
         gottago = cls._options.roles.get("default", gottago)
+    if hasattr(cls, '_options') and hasattr(cls._options, 'fields_order'):
+        fields_order = cls._options.fields_order
+        if fields_order is not None:
```
kracekumar: How about `if not fields_order` ?

schematics/transforms.py
```diff
@@ -189,6 +193,10 @@ def export_loop(cls, instance_or_dict, field_converter,
 
     # Return data if the list contains anything
     if len(data) > 0:
+        if fields_order is not None:
```
kracekumar: Same here, `if not fields_order`.

schematics/transforms.py
```diff
@@ -189,6 +193,10 @@ def export_loop(cls, instance_or_dict, field_converter,
...
+            for key in fields_order:
```
kracekumar: How about having helper function which will do this job ?

schematics/transforms.py
```diff
@@ -189,6 +193,10 @@ def export_loop(cls, instance_or_dict, field_converter,
...
+                if key in data:
+                    data.move_to_end(key)
```
kracekumar: `move_to_end` is only available in Python 3. Here is port of that in Python 2.7 `https://github.com/twitter/commons/blob/master/src/python/twitter/common/collections/ordereddict.py#L285`. But I won't support having extra dependency. If you can have logic to figure out `Py2k/Py3k`  in helper function and do proper transformation, we're good to go.

delicb:
Hi @kracekumar, thanks for your response.

I forgot about ``move_to_end`` in python 2, since I work with python 3 most of the time. I have implemented dict sorting in a way that does not use ``move_to_end`` and I have fixed code based on rest of your suggestions.

However, problem is not only ``move_to_end``, problem is that ``OrderedDict`` is not supported in python 2.6, which is supported by schematics. Not sure how to overcome that without introducing extra dependency or without having some sort of implementation of ``OrderedDict`` inside schematics itself.

I would be happy to implemenet bare minimum implementation of ``OrderedDict`` that can be used only in python 2.6, if that is something that is acceptable by you.

bintoro:
There's an `OrderedDict` implementation in `datastructures.py`.

delicb:
@bintoro, thanks, I missed that... 

Ok, now I am using ``OrderedDict`` from ``datastructures.py`` and tests for all python versions are passing.

Please let me know if you have any further suggestions.

tests/test_ordered_serialization.py
```diff
@@ -0,0 +1,58 @@
...
+        class Options:
+            fields_order = ['one', 'two', 'three']
+
+    p1 = Example({'one': 'a', 'two': 2, 'three': '3'})
+    serialized = p1.to_native()
+    comp = OrderedDict()
+    comp['one'] = 'a'
+    comp['two'] = 2
+    comp['three'] = '3'
+    assert serialized == comp
```
kracekumar: How about adding asserting returned value is `OrderedDict` ?

```
In [79]: pairs = [('a', 1), ('c', 3), ('b', 2)]
In [81]: assert dict(pairs) == OrderedDict(pairs)
```

delicb: But isn't the whole point checking if returned value is in right order? I am not sure that I understand what your proposal is... Would you mind giving an example of what would you add in this test using variable names that I used?

kracekumar: Sorry for not being verbose. I think the best way to check order is to check for return value of `items` or type of return value.

```
In [107]: serialized = OrderedDict()

In [108]: serialized['one'] = 'a'

In [109]: serialized['two'] = 2

In [110]: serialized['three'] = '3'

In [111]: order = [('one', 'a'), ('two', 2), ('three', '3')]
In [113]: assert serialized.items() == order

In [114]: assert dict(serialized.items()) == order
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-114-177698dac240> in <module>()
----> 1 assert dict(serialized.items()) == order

AssertionError:

In [115]: assert dict(serialized) == serialized

In [116]:
```

The problem with above approach is if `to_native` returns `dict` still above test will pass, rather than that checking the output of `.items` is fool proof. Depending on the `ordereddict` or `dict` output of items are different.

```
In [117]: serialized.items()
Out[117]: [('one', 'a'), ('two', 2), ('three', '3')]

In [118]: dict(serialized).items()
Out[118]: [('one', 'a'), ('three', '3'), ('two', 2)]

```

delicb: Thanks for clarification, I think I get it now. I have updated tests to compare pairs.

bintoro:
I'm puzzled by this patch.

Is there an actual use case for specifying an order other than the one in which the fields appear in the class definition?

Even if there is a use case, surely it would be better to sort the items once when the application starts, not 1000 times for an export job of 1000 objects, no?

The code also totally breaks unless all fields are present in `fields_order`.

So yeah, some changes are definitely in order (heh). What I'm asking is if this feature is even needed.

PR #258, opened 2014-11-16
Handle mocking of ModelType and ListType fields.

amibiz:

PR #306, opened 2015-06-02
Get mock object for ModelType and ListType

kaiix:
Make minor changes and add some test cases for #258


bintoro:
Hey @kaiix, would you mind resolving the conflicts so I can merge this? Thanks!

schematics/types/compound.py
```diff
@@ -151,6 +158,21 @@ def __init__(self, field, min_size=None, max_size=None, **kwargs):
...
+
+        if min_size > max_size:
+            message = u'Minimum list size is greater than maximum list size.'
+            raise MockCreationError(message)
+
+        random_length = get_value_in(min_size, max_size)
+
+        if isinstance(self.field, ModelType):
+            return [self.field.model_class.get_mock_object(context) for _ in xrange(random_length)]
+        else:
```
bintoro: Hmm, are you calling `self.field.model_class.get_mock_object()` directly just to avoid the extra function call via `ModelType._mock()`? If so, then I'd be inclined to always call `self.field._mock()` for simplicity.

kaiix:
@bintoro conflicts resolved, and use `field._mock()` to get a mock object

Issue #312, opened 2015-07-02
Model __init__ doesn't accept kwargs


jordonwii:
Line 229 of models.py says the following:
`Initial field values can be passed in as keyword arguments to ``__init__`` to initialize the object with. `

But Model.__init__ doesn't take **kwargs, and attempting to initialize the following model with kwargs raises a TypeError. 
```python
class MyModel(Model):
  my_field = StringType()

m = MyModel(my_field="test")
# => TypeError: __init__() got an unexpected keyword argument 'my_field'
```

jordonwii:
I'm willing to add a patch with tests for this if this is functionality you want to have.

kracekumar:
Sure. Please go ahead

bintoro:
I'm not sure if this is a good idea, as arbitrary keyword args have the potential to cause conflicts with any new reserved arguments that might be introduced in the future.

kracekumar:
@bintoro Do you think, `kwargs` should check for existing field name, if not throw an exception ?

```python
class V(Model):
   id = IntType()

v = V(id=1, foo=2)
# Should throw an error `foo` is unknown argument ?
```

bintoro:
Here's what I mean. Assume a program populates a field called `foo` via `**kwargs`. Later, we decide to include a named parameter `foo` in the `Model.__init__` of a future Schematics version. The program will then break, as the value passed for `foo` will be assigned to the "official" parameter instead of being interpreted as a field value.

For the above reason, and because there seems to be no tangible use case (we already have `raw_data` for precisely this purpose), I'm not in favor of adding this feature.

jordonwii:
@bintoro That argument makes sense to me. 

In that case, the comment I referenced in the original issue should probably be removed.

kontrafiktion:
I just discovered Schematics and the first thing I thought: "why don't they use keyword args" and that's how I found this issue.

> I'm not sure if this is a good idea, as arbitrary keyword args have the potential to cause conflicts with any new reserved arguments that might be introduced in the future.

I would propose that a new reserved arguments is considered an incompatible change and introduce a "NewModel" class. Then you will have (i) no conflict with existing code and (ii) any code that wants to use a new non-default value for this argument has to be changed anyway.

bintoro:
[...]

PR #313, opened 2015-07-02
Update Model to accept kwargs to its constructor as initial field values. These will override dict raw_data values with same key.

jordonwii:
Addresses #312 

I opted to override values provided in a raw_data dict for simplicity. We could do one of the following instead if preferred:

1. the inverse behavior
2. raise ConversionError if raw_data contains a key provided as a kwarg.

### DIFF
--- a/docs/usage/models.rst
+++ b/docs/usage/models.rst
@@ -44,6 +44,13 @@ And remember that ``DateTimeType`` we set a default callable for?
   >>> wr.taken_at
   datetime.datetime(2013, 8, 21, 13, 6, 38, 11883)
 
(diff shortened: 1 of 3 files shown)


jhorman:
+1

lkraider:
-1
This is unsafe if data contains keys used by the constructor.

kstrauser:
-1

I agree with @lkraider. This is pretty and useful in the common case, but provides too much opportunity for foot-shooting. Suppose a model defines `strict` as a required field. This syntax would make it impossible to create an instance like `foo = Model(strict='sure is!')`, so you'd have to use the "old" syntax for instantiating only that one model while others could use the new syntax.

PR #318, opened 2015-07-12
Propagate `strict` overrides through nested models

kstrauser:
This allows model-instantion-type overrides of strictness settings so that use cases like this work:

``` python
    class Inner(Model):
        expected_key = StringType()

    class Outer(Model):
        """docstring"""
        inner = ModelType(Inner)

    doc = {
        'inner': {
            'expected_key': 'expected value',
            'rogue_key': 'unexpected value',
        },
    }

   print Outer(doc, strict=False).inner.expected_key
```

The default behavior of honoring the models' configured strictness is maintained, but explicitly setting `strict=[value]` overrides that behavior in expected ways.

### DIFF
--- a/docs/usage/importing.rst
+++ b/docs/usage/importing.rst
@@ -53,6 +53,64 @@ data should be converted to model instances.
   u'Dillinger Escape Plan'
 
 
+On strictness
+=============
+
+When we're creating model instances from external data, we usually
+want to ensure that there are no extra, unexpected fields. Schematics
+calls these "rogue fields". By default, models raise a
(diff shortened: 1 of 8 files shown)


bintoro:
#359 fixed the propagation issue, but it tentatively removed the `ModelType.strict` option since I wasn't sure if there's an actual use case for it or if it was introduced as a half-baked way to get some control over the strictness of embedded models (since the propagation didn't work).

If there is indeed a true need for `ModelType.strict`, then let's leave this PR open until those tests pass too.

kstrauser:
Yay! \o/

PR #319, opened 2015-07-13
Remove mock from install_requires

breerly:
`mock` is only being used in tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,6 @@ setup(
     },
     install_requires=[
         'six>=1.7.3',
-        'mock==1.0.1',
     ],
     tests_require=tests_require,
 )


setup.py
```diff
@@ -51,7 +51,6 @@ def run_tests(self):
     },
     install_requires=[
         'six>=1.7.3',
-        'mock==1.0.1',
```
blampe:
```
test_requires=[
    'mock'
]
```

breerly: They already have it specified elsewhere, and tests are passing.

kracekumar: https://github.com/schematics/schematics/blob/development/tox.ini, https://github.com/schematics/schematics/blob/development/requirements-testing.txt

blampe: :+1:

kstrauser:
Works for me.

We need to lunch soon.

PR #320, opened 2015-07-14
Removed the doc section about the project's fork.

jmsdnns:
This section of the docs references some forking history that is no longer visible on the project's page.


jmsdnns:
Thanks @kracekumar!

PR #321, opened 2015-07-16
Do not compile regex

xealot:
Since this is only used during validation I don't see a strong reason to compile the regex, Python will do this on the fly and it will also enable deepcopy to work which is a requirement for subclassing models.


kracekumar:
As per this thread [is-it-worth-using-pythons-re-compile](https://stackoverflow.com/questions/452104/is-it-worth-using-pythons-re-compile), there isn't great performance benefit. So will merge the respective PR.

devnoop:
Same issue on MultiligualStringType.

PR #329, opened 2015-08-30
Add support of list on top of model

mlyundin:
Hello,

Please review my changes. Possible I have missed something as It is new project for me. 
I have used schematics for json validation. json supports list as top level object. I have found that schematics does not allow to describe Model with list on top level.

### DIFF
--- a/schematics/transforms.py
+++ b/schematics/transforms.py
@@ -3,7 +3,7 @@
 import collections
 import itertools
 
-from six import iteritems
+from six import iteritems,itervalues
 
 from .exceptions import ConversionError, ModelConversionError, ValidationError
 from .datastructures import OrderedDict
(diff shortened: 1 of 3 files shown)


jmsdnns:
I'm sorry, it is not clear to me what problem this solves.

If I look at [your test here](https://github.com/mlyundin/schematics/commit/17d7e7c630f206ca69c6bcc114b93f7c874b0340#diff-777b4783008ee8154fcbe000532ca753R219) I see that the `Group` instance is created without passing in the variable name `users`.

This feels like syntactic sugar that comes from making the structure of the data _less_ clear. I do not want to merge in code that makes things less clear as it goes against the zen of Python.

We appreciate your hard work! We hope this doesn't deter you from contributing more ideas!

Issue #307, opened 2015-06-03
Return index of invalid objects


netcriptus:
So, in my models I define a `user`, and a `user` has many `addresses`, which are defined in their own class. So when mapping the model to a schematics class, it becomes something like this

```
class UserResource(BaseSchema):

    """User resource."""

    __entity__ = User
    user_addresses = ListType(ModelType(UserAddressResource))
```

Here's the problem: imagine I have a user with 5 addresses, but 2 of them are invalid. Currently, schematics returns something like `UserAddresses: [{'street': ['Not a well formed street name.']}, {'street': ['Not a well formed street name.']}]`.

It would be great to have the index of those invalid objects in the list. My only current option is to run the validation again on each object until I find the invalid ones. But this is not practical, considering it will be doing the same work twice, and models usually have more than one relationship, so it's linked to many other models.

di:
Seems like a duplicate of https://github.com/schematics/schematics/issues/242.

netcriptus:
I think they are really close, but not duplicated. In that case, it was a list of values, and from there on schematics returns which values are invalid. In this case it is a list of objects, with its own validations.

So, in an object with 5 addresses, it is not ideal just to know that `street` is an invalid field, I need to know which of the 5 addresses has the invalid field.

bintoro:
This indeed looks like a dupe of #242. It makes no difference whether the `ListType` contains integers or models, the goal is to get the index of the faulty item within the list.

bintoro:
The proposed solution is described in #369 and implemented in #374, if anyone would like to comment on the new error message structure.

PR #331, opened 2015-09-30
Change ListType validation messages to be a dict of index to message

chase-seibert:
Currently, ListType validation error messages do not tell you WHICH item
in the list was invalid; just that one of them was. This is problematic
for showing validation errors in a UI; you typically want to show the
message next to a specific field.

I believe that a dict of item index to one or more errors (in a list) is
the most usable format for UI developers. I'm also open to returning a
list of dicts, where the position in the list matches the position in
the input; though that would often necessitate empty placeholders.

Addresses: https://github.com/schematics/schematics/issues/307

Also, fix assertions on a number of tests for validate messages:

These pytest.raises blocks stop executing when the ValidationError is
raised, meaning that any code underneath that in the block is not
executed. This included a few assertions on the actual validate
messages.

### DIFF
--- a/schematics/types/compound.py
+++ b/schematics/types/compound.py
@@ -189,12 +189,13 @@ class ListType(MultiType):
             raise ValidationError(message)
 
     def validate_items(self, items):
-        errors = []
-        for item in items:
+        errors = {}
+        for index, value in enumerate(items):
             try:
-                self.field.validate(item)
+                self.field.validate(value)
             except ValidationError as exc:
-                errors.append(exc.messages)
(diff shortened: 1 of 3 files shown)


lkraider:
I agree with the rationale here, patch seems straightforward enough.

bintoro:
I like this idea a lot. Getting the index of the offending item is clearly useful, and this will also make the error structures consistent. Right now, `ListType` gives a list of errors, while the other compounds give a dictionary.

However, since this changes the structure of the error message object, I think it's best to target this for v2.

PR #333, opened 2015-10-08
support arbitrary kwargs to a field

smcclstocks:
This will allow us to document additional domain specific
attributes on the schema & potentially use them for certain
things.

An example would be a type attribute of `unique=True` where
all fields could be enumerated for unique attributes & auto
generate database constraints.

### DIFF
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -175,7 +175,7 @@ class BaseType(TypeMeta('BaseTypeBase', (object, ), {})):
 
     def __init__(self, required=False, default=None, serialized_name=None,
                  choices=None, validators=None, deserialize_from=None,
-                 serialize_when_none=None, messages=None):
+                 serialize_when_none=None, messages=None, **kwargs):
         super(BaseType, self).__init__()
         self.required = required
         self._default = default
@@ -184,6 +184,9 @@ class BaseType(TypeMeta('BaseTypeBase', (object, ), {})):
             raise TypeError('"choices" must be a list or tuple')
         self.choices = choices
         self.deserialize_from = deserialize_from
+
+        for key, val in kwargs:
+            setattr(self, key, val)
 
         self.validators = [functools.partial(v, self) for v in self._validators]
         if validators:
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -185,7 +185,7 @@ class BaseType(TypeMeta('BaseTypeBase', (object, ), {})):
         self.choices = choices
         self.deserialize_from = deserialize_from
 
-        for key, val in kwargs:
+        for key, val in kwargs.items():
             setattr(self, key, val)
 
         self.validators = [functools.partial(v, self) for v in self._validators]
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -293,3 +293,8 @@ def test_boolean_to_native():
     for bad_value in ['TrUe', 'foo', 2, None, 1.0]:
         with pytest.raises(ConversionError):
             bool_field.to_native(bad_value)
+
+def test_field_type_kwargs():
+    field = StringType(required=True, unique=True)
+
+    assert field.unique == True


smcclstocks:
if you don't like this because of potential reserved keyword conflicts then perhaps a raw_data equivalent for types like there is on models would be acceptable? If you'd prefer that, I can generate a new pull request.

edc:
I am not sure this is a good idea. If a Field subtype needs a special option, the subtype should implement that in its `__init__`. It's better to be explicit because the method signature would be more helpful and it catches typos in instantiation calls such as `IntType(requried=True)`.

From Zen of Python:

> Explicit is better than implicit.

smcclstocks:
My goal is to actually be very explicit when defining my models. The example I give of unique=True for instance would allow me to view a model & immediately know which attrs have unique constraints on them at a glance. Or maybe something like lowercase=True, index=True, read_only=True, deprecated=True, api_version=[1, 2], etc.

I didn't want to have to subclass every single type to be able to apply universal types attributes. Perhaps a dictionary of custom attributes would be allowed other than kwargs so folks can more trivially extend types.

Something like

foo = StringType(required=True, custom={'index': True, 'api_version': [1, 2]})


I don't care what the name of the attribute is but the idea would be super helpful & it would be extending the BaseType parameters by only 1 known attribute & we won't have to sweat the collision issues. Does that make sense? If still opposed I'll just maintain my fork going forward & that's cool too.

smcclstocks:
If a custom attribute seems acceptable then tell me what the name of the attr should be & I'll make a new pull request with tests.

edc:
I understand your use case. I think a cleaner approach is simply leave the types alone, and simply use a utility function to attach additional annotations to types.

```python
def annotate(field, **kwargs):
    assert not any(key in field.__dict__ for key in kwargs)  # avoid collision
    field.__dict__.update(kwargs)
    return field

class TestModel(Model):
    foo = annotate(StringType(required=True), index=True, api_version=[1, 2])
```

If you don't like the syntax, there is also a fancier way:

```python
def a(field_cls):
    allowed_args = list(itertools.chain(*[
        inspect.getargspec(cls.__init__).args[1:]
        for cls in field_cls.__mro__ if cls.__init__ != object.__init__
    ]))
    def ctor(**kwargs):
        init_kwargs = {k: v for k, v in kwargs.iteritems() if k in allowed_args}
        field = field_cls(**init_kwargs)
        annotations = {k: v for k, v in kwargs.iteritems() if k not in allowed_args}
        field.__dict__.update(annotations)
        return field
    return ctor

class TestModel(Model):
    foo = a(StringType)(required=True, index=True, api_version=[1, 2])
```

smcclstocks:
You're a smarter individual than I so whatever you think. All I want is a convenient way to define attrs for any type & I don't require they be direct properties on the object (instance or not). Simply passing in a context or custom that could be anything: dict, list, object, str, set  ... where type safety, naming, etc is up to the user as is enforcement. No collisions & really simple, all the API would need to expose is one additional attribute that is available on the type's object.

I'm a real dim bulb because I can hardly make sense of what you've done.

bintoro:
I think this is a reasonable request.

I'm just wondering if it would be better to put these ad-hoc attributes in their own container instead of on the object itself, à la `somefield.attrs.custom_attribute`. This way there would never be any namespace conflicts, i.e. `somefield.foo` and `somefield.attrs.foo` could happily coexist.

edc:
I apologize if my earlier comments were offensive in any way.

> if it would be better to put these ad-hoc attributes in their own container

I still think there is no need to modify Schematics itself for this use case, but if others agree the request is reasonable, I agree that it's better to have a container attribute, like `annotations`.

bintoro:
@edc They didn't seem offensive to me.

I forgot to mention that the ad-hoc attributes should absolutely be specified in a dictionary, as opposed to blindly taking any kwargs and turning them into properties.

As for the name, `annotations` seems fine semantically, I just dislike its length.

smcclstocks:
Having them accessible on a container is fine by me as well. Visually I like `field_type.attr.custom_attr` over how my proposal would access the attribute `field_type.attr['custom_attr']`. So I like your suggestion!

smcclstocks:
Oh & definitely not offensive. I'm seriously not that good at this stuff.

jmsdnns:
I don't agree with putting this login inside schematics. I believe there is a simple way of doing this using regular python and it is already done elsewhere in the codebase.

The basic idea is that your `__init__` method looks for specific keys in `**kwargs`. If it finds them, you can get it with `kwargs.pop('key')` and also remove it from the list of things passed to `super(...)`.

If a keyword argument you didn't expect is passed in, `super(...)` will be the one that notices it and throws an error.

https://github.com/schematics/schematics/blob/development/schematics/types/compound.py#L139

This was good work but I'm sorry I am going to close this.

smcclstocks:
Sorry about responding to a closed pull-request but I don't know how to follow up with questions.. let me know if there is a better way!

Is your guidance to sub-class every native schematics field type that I may want to add custom attributes to? So if I want a `description` attribute on all of my models fields that may be used to generate jsonschema docs for example I'd need to sub-class every type that I'd use to implement the logic you're talking about?

Something like:

```
class User(models):
  first_name = StringTypeCustom(description='First Name', required=True)
  age = IntTypeCustom(description='Age in units of years', min_value=21)
  registered = BooleanTypeCustom(description='Currently registered in the system', default=False)
```


In the example above, I'd sub-class all of those native types & override the constructor to do the logic you mentioned. Is that what you're talking about? Thanks for following up!

jmsdnns:
No worries, I think this _is_ the right place. And I think I didn't quite grok something you mentioned fully until I saw your example above.

I guess in this case, I see the value in your description field perhaps enough that Schematics itself might benefit from it. There was a time far back in its history when it _did_ support JSONSchema and this was cool.

I think your code speaks to a general concept of annotations. The description is itself meta data and not really adjusting functionality, so I don't think keywords is the right place for that. But perhaps a keyword `annotations` could be added to BaseType and one of the annotation types could then be "description".

I'm thinking out loud here, btw. Feel free to throw more ideas in.

smcclstocks:
I like the idea of allowing for domain specific freedom on whatever this keyword is. Honestly bro, the whole kwargs pull request I submitted I knew wouldn't fly with you guys cause of clobbering other future keywords for one but I wanted to get the use case out there & I'm not smart enough to do much else lol!

I'd hate to turn around & request expanding the API everytime I want an additional 'meta' argument like `index` for instance or `deprecated` which could be applied to any of my models fields potentially on any type of field so sub-classing would be a real bitch. The ideas the other guys had of anchoring the arbitrary properties on some meta type keyword was really simple & elegant I thought & now that you know the use case do you think so too?

My only request is it be easy when defining the field types properties. How it best fits in the code base & aligns with y'alls preferences - I'm all about doing what you guys think is best.

bintoro:
I still think this is a reasonable idea, as long as
* the arguments are passed in a dictionary, not as arbitrary keyword args
* the annotations are accessible on the field via a dedicated member like `annotations` or `attrs`.

jmsdnns:
OK, so if I try to imagine one way of expressing this idea as a Schematics model, I get here. What do you think?

```python
class Person(Model):
    first_name = StringType(required=True,
                            annotations={
                                'description': 'FirstName',
                                'as_sql_col': 'NUMBER',
                            })
    age = IntType()
```

Also, I added another type of annotation, just to capture some of what might be possible if we get annotations done right.

jmsdnns:
Haha. My `as_sql_col` value says NUMBER for a `StringType`. Github's already emailed everyone, so no point editing it... :)

smcclstocks:
It could be helpful to have a `get_types_by_annotation` method on the model to. So to keep your string -> number conversion example alive & well :)  ... something like:

<array of fields> = model.get_types_by_annotation('as_sql_col', 'NUMBER')


`get_types_by_annotation` can obviously be anything you guys want but having it as a public method could be super useful.

smcclstocks:
maybe `get_fields_by_annotation` would have been better in the example above to get the idea across.

chadrik:
I just wanted to jot down to suggestions.
First, I prefer the name `metadata` over `annotations` as I think it's more familiar and therefore self-explanatory.

I hesitate to add the second idea because it may be overkill, but here it is:

```python
class Metadata(Model):
    title = StringType()
    description = StringType(required=True)

class Person(Model):
    class Options:
        metadata_model = Metadata

    first_name = StringType(
        required=True,
        metadata=dict(
            title='First Name',
            description="The person's first name")
        )
    age = IntType(        
        metadata=dict(
            description="The person's age",
            as_sql_col='NUMBER')
        )
```

Basically, it's an optional mechanism to validate metadata.  I'm mostly interested in asserting the existence of required metadata, but the complete model could be useful in cases where the schema itself needs to be transmitted.

A simpler alternative (and therefore probably better):

```python
class Person(Model):
    class Options:
        required_metadata = ['description']
[...]
```

This is just an idea to ponder, take it or leave it.  

Anyway, I would love to see this issue re-opened, renamed to "Support arbitrary field metadata", and completed.  This should be really really easy, right?  If no one wants first dibs, I'll go ahead and make a PR (for the core feature, I'll leave out the metadata validation for now).

Issue #330, opened 2015-09-30
Model that contains field of of its type (same ModelType as itself)


mjrk:
See the following schema

    class Expression(Model):

        expressions = ListType(ModelType(Expression))

Unfortunately, this is not possible as during class definition `Expression` is unknown. I could not find how to do this in the docs, searching `schematics/models.py` doesn't seem this is implemented. However, maybe I missed something as it seems a common use case to me.

IMHO this could be implemented in the `ModelMeta` class after the `klass` creation (see models.py). Could it work to set the field value to `klass` in those cases? Either with a special type `SelfType` or some DSL.

Or we could add a more explicit way to add custom fields - e.g. a method which's return value is used to extend the `klass`' -fields after initialization.

Suggestions?

lkraider:
I went with the second option and added an `append_field` method to the Model:
https://github.com/nKey/schematics/blob/0ea74550b695ac962aab1828f63f44b909dfcdd7/schematics/models.py#L130

bintoro:
It would certainly be neat to be able to specify self-referential fields right there in the model definition.

Any opinions on what the syntax should look like?

```python
class Node(Model):
    
    children = ListType(ModelType()) # no argument
```

```python
class Node(Model):
    
    children = ListType(ModelType(SelfRef)) # some placeholder object
```

```python
class Node(Model):
    
    children = ListType(ModelType('Node')) # the name of the model
```

mjrk:
Personally, I like 

```python
class Node

    children = ListType(ModelType('Node'))
```

because it's the most explicit and you find all references with one search. On the other hand, it might make people believe they can use strings also for other models. Still +1 for it.

Using a placeholder (which I'd like second most) IMHO `schematics.types.compound.Self` seems straight forward.

Thanks a lot for your answer and great work btw.!

bintoro:
@mjrk The syntax using the name of the model was my preference as well.

An implementation is now up for review as PR #335.

bintoro:
This now exists in `development` and will be part of the next release (likely v1.2.0).

mjrk:
Oh, with the work on v2.0.0 - will there be a version v1.2.0? Would be awesome

bintoro:
@mjrk I considered making an interim release, but the thing is, there's really no specific point in history where it would clearly make sense. The road from v1.1.1 to v2.0.0 involves a series of small backward-incompatible changes (mainly for users who have implemented their own types), and by far the best way to deal with them is all at once.

[...]

PR #335, opened 2015-10-13
Add support for self-referential ModelType fields

bintoro:
Implements the feature requested in #330.

Commit message:

ModelType fields can now be initialized with a string consisting of the name of the containing model. These will be turned into actual class references once the class has been created.

### DIFF
--- a/schematics/models.py
+++ b/schematics/models.py
@@ -7,7 +7,6 @@ import sys
 from six import iteritems
 from six import iterkeys
 from six import add_metaclass
-from six import string_types as basestring
 
 from .types import BaseType
 from .types.serializable import Serializable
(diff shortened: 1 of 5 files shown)


bintoro:
I added another commit that takes the late-stage field setup code that had accumulated in `ModelMeta` and places it in the relevant type classes. Monkey-patching individual properties inside field instances shouldn't be the job of `ModelMeta`.

This also simplifies the code a bit, as recursive setup of nested fields now happens automatically for all types descending from `MultiType`.

edc:
Code looks great! Do we want to add the same support to `PolyModelType`?

bintoro:
@edc Absolutely! I just didn't get around to it yet.

Adding this feature has led me to consider an intermediate class under `MultiType` from which both `ModelType` and `PolyModelType` would inherit. They do share some logic.

mjrk:
great!

And +1 one for introducing the _setup method (that btw. offers a great hook for customization)

PR #336, opened 2015-10-15
Don't evaluate serializable when accessed through class

bintoro:
Consider:

```python
class M(Model):

    f = StringType()

    @serializable
    def s(self):
        return 'foo'
```

Since `M.f == M._fields['f']`, I would expect `M.s == M._serializables['s']` to be true as well.

But accessing `M.s` currently results in `M.s()` to be evaluated as if it were a classmethod. I suspect this is not by design. Does anyone else see a reason why this should be so?


jmsdnns:
I dig this. Keeps everything consistent in a meaningful way.

PR #340, opened 2015-10-19
First draft of PolyType

jmsdnns:
I wanted to write code that looks like this following:

```python
class Foo(Model):
    bla = ListType(PolyType([IntType(), StringType(min_length=3)]))

f1 = Foo({'bla': [(IntType, 5), (StringType, 'foo')]})
f2 = Foo({'bla': [5, 'foo']})
```

And I wanted the output to look like this:

```
>>> f1.to_native()
{'bla': [5, u'foo']}
>>> f2.to_native()
{'bla': [5, u'foo']}
```

I found `schematics.types.compound.PolyModelType`, but it only works with models and I wanted multiple choices at the type level instead, thus `PolyType`.

This code depends on [this PR](https://github.com/schematics/schematics/pull/339) which I merged in an hour ago. It's how I found the issue.

Here is a complete gist of working code using this patch and latest development branch of Schematics.

```python
from schematics.types import BaseType, StringType, IntType, PolyType
from schematics.types.compound import MultiType, ListType


class Foo(Model):
    bla = ListType(PolyType([IntType(), StringType(min_length=3)]))


f1 = Foo({'bla': [(IntType, 5), (StringType, 'foo')]})
print 'Calling f1.to_native'
print f1.to_native()

f2 = Foo({'bla': [5, 'foo']})
print 'Calling f2.to_native'
print f2.to_native()
```

_Note: One can edit the `min_length` argument to see that errors are thrown by instances of types which can each have a unique config_

### DIFF
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -885,3 +885,100 @@ class MultilingualStringType(BaseType):
             if self.locale_regex is not None and self.locale_regex.match(locale) is None:
                 raise ValidationError(
                     self.messages['regex_locale'].format(locale))
+
+
+class PolyType(BaseType):
+
+    def __init__(self, type_classes, **kwargs):
+        if isinstance(type_classes, (list, tuple)):
+            self.type_classes = tuple(type_classes)
+            allow_subclasses = False
+        else:
+            raise Exception("The first argument to PolyType.__init__() "
+                            "must be an iterable of types.")
+
+        validators = kwargs.pop("validators", [])
+        self.strict = kwargs.pop("strict", True)
+        self.claim_function = kwargs.pop("claim_function", None)
+        self.allow_subclasses = kwargs.pop("allow_subclasses", allow_subclasses)
+
+        def validate_type(type_instance):
+            type_instance.validate()
+            return type_instance
+
+        BaseType.__init__(self, validators=[validate_type] + validators, **kwargs)
+
+    def __repr__(self):
+        return object.__repr__(self)[:-1] + ' for %s>' % str(self.type_classes)
+
+    ###
+    ### Type Inspection
+    ###
+
+    def is_allowed_type(self, type_instance):
+        if self.allow_subclasses:
+            if isinstance(type_instance, self.type_classes):
+                return True
+        else:
+            # The type matches, but validation will catch error in it's shape
+            if type_instance.__class__ in [cls.__class__ for cls in self.type_classes]:
+                return True
+        return False
+
+    def find_type(self, data):
+        chosen_class = None
+        
+        if self.claim_function:
+            return self.claim_function(self, data)
+        else:
+            candidates = self.type_classes
+            if self.allow_subclasses:
+                candidates = itertools.chain.from_iterable(
+                                 ([m] + m._subclasses for m in candidates))
+                
+            matching_classes = []
+            for cls in candidates:
+                ### Find match by attempting validation
+                try:
+                    if isinstance(data, (list, tuple)):
+                        cls.validate(data[1])
+                        chosen_class = cls
+                        break
+                    else:
+                        cls.validate(data)
+                        chosen_class = cls
+                        break
+                except Exception as e:
+                    pass
+
+        if chosen_class:
+            return chosen_class
+        else:
+            raise Exception("Input for polymorphic field did not match any model")
+
+    ###
+    ### Converters
+    ###
+
+    def to_something(self, value, get_converter, context=None):
+
+        if value is None:
+            return None
+        
+        if self.is_allowed_type(value):
+            return value
+        
+        type = self.find_type(value)
+        converter = get_converter(type)
+
+        if isinstance(value, (list, tuple)):
+            return converter(value[1], context=context)
+        else:
+            return converter(value, context=context)
+
+    def to_primitive(self, value, context=None):
+        return self.to_something(value, lambda t: t.to_primitive, context=context)
+    
+    def to_native(self, value, context=None):
+        return self.to_something(value, lambda t: t.to_native, context=context)
+    


edc:
Would [UnionType](https://en.wikipedia.org/wiki/Union_type) be a better name for this?

jmsdnns:
Hmm. Makes sense to me.

If we change this, I think we should probably also change `PolyModelType` to be `UnionModelType` too, for consistency.

Thoughts?

edc:
I think `PolyModelType` is a fine name, because in most use cases it is used with a root class and its descendants and aims at taking advantage of polymorphism in a classic OOP way. "member" types in a `PolyModelType` often form a tree structure, while structure of those of a union type is flat.

I do think we can eventually deprecate the use of [passing an iterable of types](https://github.com/schematics/schematics/blob/dac37186be514b6091f451e8ca6eb24778515378/schematics/types/compound.py#L321-L324) to `PolyModelType` and move those use cases to `UnionType`, i.e. write `UnionType([TruckType(), BoatType()])` instead of `PolyModelType([TruckType(), BoatType()])`.

bintoro:
Actually, polymorphism in general refers to a variety of situations where a range of types can be substituted for one another. "Polymorphism in a classic OOP way" probably means subtyping, which is not the same as polymorphism, just a special case of it. So there's nothing wrong with `PolyType`, I think.

bintoro:
> I do think we can eventually deprecate the use of passing an iterable of types to `PolyModelType` and move those use cases to `UnionType`, i.e. write `UnionType([TruckType(), BoatType()])` instead of `PolyModelType([TruckType(), BoatType()])`.

`PolyType/UnionType` takes an iterable of *types*, while `PolyModelType` is already a variant of `ModelType` that takes an iterable of *models*.

```PolyModelType([A, B, C])```

is equivalent to

```PolyType([ModelType(A), ModelType(B), ModelType(C)])```.

File: schematics/types/base.py [hunk 1]
```diff
@@ -885,3 +885,100 @@ def validate_regex(self, value):
             if self.locale_regex is not None and self.locale_regex.match(locale) is None:
                 raise ValidationError(
                     self.messages['regex_locale'].format(locale))
+
+
+class PolyType(BaseType):
+
+    def __init__(self, type_classes, **kwargs):
+        if isinstance(type_classes, (list, tuple)):
+            self.type_classes = tuple(type_classes)
```
bintoro: The `type_classes` member (and the corresponding `__init__` parameter) should perhaps be called `type_instances` or just `types` instead, since (unlike in `PolyModelType`) the items are not really classes.

jmsdnns:
This is the line that swayed me in the wiki page, regarding _union_ or _poly_.

_In other words, a union type definition will specify which of a number of permitted primitive types may be stored in its instances, e.g., "float or long integer"._

My implementation above lists the types specifically. We could use poly, but in this case, I believe the name Union is more specific about what is actually happening.

With respect to the models, I think can't help but feel like it's not actually correct to have the `PolyModelType`. Consider that by using that type, one can only use models and not use types. Exactly the reason I felt limited enough to create a new type. So, to go further, one _could_ put ModelType's inside the PolyType and have types or models in the list together.

bintoro:
> So, to go further, one could put ModelType's inside the PolyType and have types or models in the list together.

I would like to unify them as well. `PolyModelType` could remain as an alias, since

```python
PolyModelType([A, B, C], allow_subclasses=True)
```

is way clearer than

```python
PolyType([ModelType(A, allow_subclasses=True), ModelType(B, allow_subclasses=True), ModelType(C, allow_subclasses=True)]).
```

@jmsdnns, did you have plans to use the `allow_subclasses` parameter in `PolyType`?

In `PolyModelType` it's used to control whether model matching should be strict or if subclasses are OK. The analogous thing in `PolyType` would be to control whether subclasses of *types* are OK. If we want to implement that sort of thing, I think it should be handled at the `BaseType` level. There's no reason why such a setting should only exist inside `PolyType` and not when the type resides directly at the model level.

As an example, the `allow_subclasses` parameter for types would determine if

```python
f1 = Foo({'bla': [(MyCustomIntType, 5), (StringType, 'foo')]})
```

would be acceptable instead of

```python
f1 = Foo({'bla': [(IntType, 5), (StringType, 'foo')]}).
```

schematics/types/base.py
jmsdnns: You're totally right! Thanks.

jmsdnns:
Oh hmmmm. I didn't actually understand what `allow_subclasses` was about yet, but now I get it. That's a really cool idea. I'll add that to the PR before anything gets merged in.

I also see the point you're making about the simpler syntax. If I take it to an extreme, I think we get here, and I'm not sure what to think of this yet...

```python
x = PolyType([IntType(), PolyModelType([A, B, C], allow_subclasses=True), StringType(min_length=10)])
```

Not too bad... but still verbose. Thoughts?

bintoro:
The way I see it, the current responsibilities of `PolyModelType` would be split between `ModelType` and `PolyType`, and `PolyModelType` would only remain as an alias for its current use cases.

If someone really defines a field that accepts both an integer and a range of models, they're just going to have to type out the individual `ModelType(...)` entries...

jmsdnns:
I like the second idea best, but it's gonna take a few PR's to get there.

Specifically, I want this to work:

```python
x = PolyType([IntType(), A(), B(), StringType()])
```

bintoro:
If I understand the objective correctly, then it seems this thing would have to continually keep track of the current value's type as well.

`PolyModelType` works because the values are always model instances, and a model instance always knows which model it represents. But here, the native values can be basic datatypes that don't carry such information.

Consider this hypothetical situation:

```python
class Part(Model):
  ...
  third_party_id = PolyType([StringType(), SupplierPartCodeType()])
```

Let's say `SupplierPartCodeType` has a `to_primitive` method that always formats the exported value as `nnnn-nn-nnn`, so in this case `1223-45-678`. But that's not what happens right now:

```python
>>> part = Part({'third_party_id': (SupplierPartCodeType, '122345678')})
>>> part.to_primitive
{'third_party_id': '122345678'}
```

Even though `SupplierPartCodeType` is specified upon import, this information is lost right away. Upon export, `StringType` is tried before `SupplierPartCodeType`, and, since it matches the internal representation `"122345678"`, it gets to perform the transformation despite not being the intended type.

bintoro:
Unless I'm missing something obvious, this fully flexible `PolyType` idea would seem to require pretty fundamental changes to the way Schematics stores data.

In my previous comment I pointed out that the type information "must be stored somewhere" without fully realizing how complicated that's going to get.

See, values go in `_data`, but `_data` includes zero information about their Schematics types. This has been possible because, up until now, model definitions have been unambiguous about which type corresponds to which value.

For instance, `_data` might contain `"8.8.8.8"`. Just by looking at the value, it's impossible to tell if it's a `StringType` or an `IPv4Type`, but that's okay since the model knows.

However, once you do this, you introduce ambiguity:

```python
class Foo(Model):
  misc_items = ListType(PolyType([StringType(), IPv4Type()]))
```

When exporting the model, `ListType` iterates through `misc_items`, passes each element to `PolyType`, and tells it to serialize the value. But `PolyType` has absolutely no way of knowing whether it should delegate the serialization task to `StringType` or `IPv4Type` (unless one of them fails to validate, which cannot be relied on).

So, either the `PolyType` instance or a centralized datastore like `_data` would have to constantly maintain metadata about the intended type of every value.

Regular Python lists would go out the window, since there's no way to intercept operations on `foo_instance.misc_items` unless it's a magical Schematics list. For example, SQLAlchemy uses [InstrumentedList](http://docs.sqlalchemy.org/en/latest/orm/collections.html#sqlalchemy.orm.collections.InstrumentedList) & friends in place of built-in datatypes in order to track changes to collections.

While all of this is totally doable, it's quite an undertaking.

bintoro:
Comparable functionality was implemented in #380, so I believe this may now be closed.

PR #344, opened 2015-10-23
Fix longitude values for GeoPointType mock and add tests

wraziens:


lkraider:
Looks simple enough, maybe we can have some tests for floating point coordinates up to 6 digits (around 10cm or 4in of precision).

PR #345, opened 2015-10-28
Support Python 3.5

jonafato:
- Add py35 to list of tox environments
- Add Python 3.5 to language classifiers
- Upgrade pytest version to address issues with Python 3.5
  (https://github.com/pytest-dev/pytest/pull/801)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,13 @@ install:
   - pip install tox
   - pip install python-coveralls
 
+addons:
+  apt:
+    sources:
+      - deadsnakes
+    packages:
+      - python3.5
+
 script: tox
 
 after_success:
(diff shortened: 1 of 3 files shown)


jonafato:
The failure here seems to be related to https://github.com/travis-ci/travis-ci/issues/4990 and should resolve itself once that issue is closed.

bintoro:
Looking at the related Travis issues, it turns out the on-demand installation works if you just specify:

```
python:
  - 3.5
```

No need to install via `apt`.

6acd4e9f0ce932a0d89ed970809844d914d2749f / https://travis-ci.org/schematics/schematics/builds/91870152

bintoro:
Fixed while switching over to the new Travis infrastructure.

PR #346, opened 2015-10-28
Combining Requirements

### LINKED ISSUE #223: Travis and tox requirements should be united?
There's currently `requirements-testing.txt` (used by tox) and `dev-requirements.txt` (used by Travis). Both have the same core goal: assuring all dependencies needed for testing are installed. Is there a reason to keep them separate, or could we merge them into a single file?

arosenfeld:
Combined requirements-testing.txt and dev-requirements.txt.  Resolves  #223.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,6 @@
 include README.rst
 include setup.py
-include requirements-testing.txt
+include requirements.txt
 include tox.ini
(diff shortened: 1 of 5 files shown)


bintoro:
Seeing as Sphinx is the only difference between the two sets of requirements, I suppose this makes sense. Thanks!

PR #349, opened 2015-10-31
Fix FieldDescriptor.__delete__ to not touch model

bintoro:
Currently, `del model_instance.somefield` performs

```python
del model_instance._fields['somefield']
```

But instances don't have a `_fields` member, so this results in `somefield` to be dropped not from the instance but **from the class, affecting all instances**.

This patch changes the behavior of `FieldDescriptor.__delete__` to match that of `Model.__delitem__`, which simply resets the field's value to `None`.

If manipulating `_fields` is necessary, it should be done through the class.

### DIFF
--- a/schematics/models.py
+++ b/schematics/models.py
@@ -71,12 +71,12 @@ class FieldDescriptor(object):
 
     def __delete__(self, instance):
         """
-        Checks the field name against a model and deletes the field.
+        Checks the field name against a model and deletes the value.
         """
-        if self.name not in instance._fields:
-            raise AttributeError('%r has no attribute %r' %
-                                 (type(instance).__name__, self.name))
-        del instance._fields[self.name]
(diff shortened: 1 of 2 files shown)


bintoro:
I'm gonna go ahead and merge this, since the current behavior is very weird, and to support unset fields, there must be a way to clear a field's value. `del obj.somefield` is the obvious choice for resetting a field.

j2labs:
Good! Your point here was totally right

On Wednesday, November 11, 2015, Kalle Tuure <notifications@github.com>
wrote:

> Merged #349 <https://github.com/schematics/schematics/pull/349>.
>
> —
> <https://github.com/schematics/schematics/pull/349#event-461234413>.
>


-- 
jms dnns
<https://twitter.com/jmsdnns>

PR #350, opened 2015-10-31
export_loop cleanup

bintoro:
In `export_loop` functions,
* get rid of redundancies in conditionals
* when about to return `None`, do so explicitly.

Can someone with more historical knowledge please check if any of the removed code in `compound.py` was intended to do something but didn't because it was buggy?

Specifically:

1. In `ModelType.export_loop()` we had

  ```python
  if shaped and len(shaped) == 0 and self.allow_none():
      return shaped
  elif shaped:
      return shaped
  elif print_none:
      return shaped
  ```

  All cases covered by the first conditional are covered by the second as well. But maybe the second one was supposed to read something like:

  ```python
  elif shaped and len(shaped) > 0:
      return shaped
  ```

  But still, this would only make sense for a custom representation that's not a `dict`, yet the main `export_loop` function only returns dictionaries.

2. The `feels_empty` variable in `ListType/DictType.export_loop()` didn't serve any purpose, which I found strange.

### DIFF
--- a/schematics/types/compound.py
+++ b/schematics/types/compound.py
@@ -110,12 +110,10 @@ class ModelType(MultiType):
                              field_converter,
                              role=role, print_none=print_none)
 
-        if shaped and len(shaped) == 0 and self.allow_none():
+        if shaped or print_none:
             return shaped
-        elif shaped:
-            return shaped
-        elif print_none:
-            return shaped
+        else:
+            return None
 
 
 class ListType(MultiType):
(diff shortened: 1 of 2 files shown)


bintoro:
Regarding `feels_empty`, if this line

```python
feels_empty = shaped and len(shaped) == 0
```

was instead

```python
feels_empty = shaped is None or len(shaped) == 0
```

then it would make more sense. I've amended the commit to reflect what I think the code has been trying to accomplish.

Presumably the intent has been that `feels_empty` should be `True` for empty containers. Well, it hasn't because `bool([])` is obviously `False`. The `serialize_when_none` setting has been filtering out empty containers not because they "feel empty" but because the containers have been programmed to return `None` when `serialize_when_none` is `False`.

This is now fixed so that the containers could just as well return `[]` or `{}` and still be affected by `serialize_when_none` at the model level.

schematics/types/compound.py
```diff
@@ -196,26 +194,23 @@ def export_loop(self, list_instance, field_converter,
             if hasattr(self.field, 'export_loop'):
                 shaped = self.field.export_loop(value, field_converter,
                                                 role=role)
-                feels_empty = shaped and len(shaped) == 0
+                feels_empty = shaped is None or len(shaped) == 0
             else:
```
bintoro: Basically I was looking for the original thought behind this line.

`feels_empty` has always been `False` here because in no situation would `bool(shaped) == True` and `len(shaped) == 0` occur simultaneously.

Still looks good to me, and this is necessary preparation for #351 and the undefined feature, so merging.

Issue #235, opened 2014-06-18
Models without fields are serialized to None


magaman384:
Hello,
I'm not sure if it is bug or feature but following model will be converted to None after calling serialize:
```
class EmptyModel(Model):
    pass

model = EmptyModel({})
model.serialize() # this returns None instead of {}
```

bintoro:
Contrary to the previous comment, this is not the same as the "unset field" issue, although it's related. This here is about the serialized representation of empty model instances.

An empty instance will currently be exported as `None`. OP expected to see `{}`.

An instance is empty if
* its class doesn't define any fields (OPs example) or
* none of its fields contains a nonempty value AND `serialize_when_none=False`.

```python
class Test(Model):

    class Options:
        serialize_when_none = False

    l = ListType(IntType())
    s = StringType()
```
```python
>>> t = Test()
>>> t.l = []
>>> print(t.to_primitive())
None
```

The behavior is consistent with `ModelType` fields:


```python
class Outer(Model):
    m = ModelType(Test)
```
```python
>>> o = Outer()
>>> o.m = Test()
>>> print(o.to_primitive())
{'m': None}
```

The question is: is `None` the ideal representation for empty models?

Well, I think `None` makes perfect sense for now, and I don't know how `{}` would be an improvement.

The only argument for `{}` that I can think of is that the current behavior is not entirely consistent between `ModelType` and, e.g., `ListType`:
* `ListType` accepts both `None` and `[]`, and they're serialized as such.
* `ModelType` accepts both `None` and `Foo()`, but they're both serialized as `None`.

My take on this is that the `ModelType` behavior is actually desirable, and it's `ListType` that's acting strange. The fact that `ListType` can already accommodate two different kinds of "empty" means that, once the Undefined feature lands, there will be three kinds.

Therefore, the present issue seems to need addressing only if

* there's currently a scenario where a model or a `ModelType` field will get serialized as `{}`, as that would be inconsistent, or

* there would be some kind of semantic difference between `None` and `{}`, but I don't readily see what that might be.

bintoro:
Let me revise my previous statement.

Obviously, there is a difference between `obj.field = None` and `obj.field = ModelType(EmptyModel)`. The latter is an actual object even if it has nothing to export.

It's just that, under default settings, this is such an artificial scenario so as not to matter.

[...]

PR #351, opened 2015-10-31
Always export empty compound objects as {} / []

bintoro:
This PR changes the primitive representation of empty models from `None` to `{}` (empty `dict`) so as to make a distinction between an actual `None` and an instance that has no values to export.

It depends on #350, which fixes the filtering of empty containers during export, and addresses  #235.

Diff for the relevant commit: https://github.com/schematics/schematics/commit/912afcc02c5506290d8fe5d818207c141cc6d1e3

#### Motivation

The current behavior can lead to counterintuitive outcomes, particularly when using the `serialize_when_none` switch.

Example:

```python
class Inner(Model):
    class Options:
        serialize_when_none = False
    l = ListType(IntType())

class Outer(Model):
    m = ModelType(Inner)
```

```python
>>> instance = Outer()
>>> instance.to_primitive()
{'m': None} # as expected since instance.m hasn't been set
```

```python
>>> instance.m = Inner({'l': [123]})
>>> instance.to_primitive()
{'m': {'l': [123]}} # again as expected
```

Let's mutate the list:

```python
>>> instance.m.l.pop()
>>> instance.to_primitive()
{'m': None} # !!!
```

The `Inner` object still very much exists at `instance.m` but is reduced to `None` because it has nothing to export.

The patch changes the output in this case to `{'m': {}}`.

### DIFF
--- a/schematics/transforms.py
+++ b/schematics/transforms.py
@@ -190,11 +190,7 @@ def export_loop(cls, instance_or_dict, field_converter,
     if fields_order:
         data = sort_dict(data, fields_order)
 
-    # Return data if the list contains anything
-    if len(data) > 0 or print_none:
-        return data
-    else:
-        return None
+    return data
 
 
 def sort_dict(dct, based_on):
--- a/schematics/transforms.py
+++ b/schematics/transforms.py
@@ -178,26 +178,23 @@ def export_loop(cls, instance_or_dict, field_converter,
                 shaped = field_converter(field, value)
 
             # Print if we want none or found a value
-            if shaped is None and allow_none(cls, field):
-                data[serialized_name] = shaped
-            elif shaped is not None:
-                data[serialized_name] = shaped
-            elif print_none:
+            if shaped is not None \
+              or allow_none(cls, field) \
+              or print_none:
                 data[serialized_name] = shaped
 
         # Store None if reqeusted
-        elif value is None and allow_none(cls, field):
-            data[serialized_name] = value
-        elif print_none:
+        elif allow_none(cls, field) or print_none:
             data[serialized_name] = value
 
+    if fields_order:
+        data = sort_dict(data, fields_order)
+
     # Return data if the list contains anything
-    if len(data) > 0:
-        if fields_order:
-            return sort_dict(data, fields_order)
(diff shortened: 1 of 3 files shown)


bintoro:
I added a couple of commits here that simplify things even further.

Previously, the responsibility for dropping unwanted "empty" fields has been split between the containing object and the affected field itself.

For instance, `ListType.export_loop()` has ended with this:

```python
# Return data if the list contains anything
if len(data) > 0 or self.allow_none() or print_none:
    return data
else:
    return None
```

This is because the outer field hasn't been checking for emptiness correctly.

#350 fixed the `feels_empty` check in compound types, and now the same logic has been copied over to the main `export_loop` implementation. The emptiness check is now consistently performed on the container side, so a field doesn't have to concern itself with whether it should be dropped from its *parent's* representation.

So the above logic becomes just:

```python
return data
```

The whole `serialize_when_none` thing should now be way easier to understand, since the checks aren't happening all over the place.

All tests pass.

j2labs:
seems sensible to me :)

On Wed, Nov 11, 2015 at 9:38 AM, Kalle Tuure <notifications@github.com>
wrote:

> I added a couple of commits here that simplify things even further.
>
> Previously, the responsibility for dropping unwanted "empty" fields has
> been split between the containing object and the affected field itself.
>
> For instance, ListType.export_loop() has ended with this:
>
> # Return data if the list contains anythingif len(data) > 0 or self.allow_none() or print_none:
>     return dataelse:
>     return None
>
> This is because the outer field hasn't been checking for emptiness
> correctly.
>
> #350 <https://github.com/schematics/schematics/pull/350> fixed the
> feels_empty check in compound types, and now the same logic has been
> copied over to the main export_loop implementation. The emptiness check
> is now consistently performed on the container side, so a field doesn't
> have to concern itself with whether it should be dropped from its
> *parent's* representation.
>
> So the above logic becomes just:
>
> return data
>
> The whole serialize_when_none thing should now be way easier to
> understand, since the checks aren't happening all over the place.
>
> All tests pass.
>
> —
> <https://github.com/schematics/schematics/pull/351#issuecomment-155799450>
> .
>



-- 
jms dnns
<https://twitter.com/jmsdnns>

PR #355, opened 2015-11-03
Support returning a sanitized value from validators

bintoro:
The docs imply that validators may return a modified value to replace the original, but this is not true (#206). This PR adds support for return values from both type-level and model-level validators.

Since most validators today likely return an implicit `None`, a validator that is *really* going to return a modified value must be specially designated as such. This is done by decorating the validator with `@returns` (defined in `validate.py`).

Custom types that override `BaseType.validate()` and fail to return a value will stop working (fields will be set to `None` upon validation). But overriding `validate()` is probably rare, and the docstring clearly says to "return a clean value".

### DIFF
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -236,7 +236,9 @@ class BaseType(TypeMeta('BaseTypeBase', (object, ), {})):
 
         for validator in self.validators:
             try:
-                validator(value)
+                clean_value = validator(value)
+                if getattr(validator, 'returns', False) is True:
+                    value = clean_value
             except ValidationError as exc:
                 errors.extend(exc.messages)
 
@@ -245,6 +247,8 @@ class BaseType(TypeMeta('BaseTypeBase', (object, ), {})):
 
         if errors:
             raise ValidationError(errors)
+
+        return value
 
     def validate_required(self, value):
         if self.required and value is None:
--- a/schematics/types/compound.py
+++ b/schematics/types/compound.py
@@ -25,7 +25,9 @@ class MultiType(BaseType):
 
         for validator in self.validators:
             try:
-                validator(value)
+                clean_value = validator(value)
+                if getattr(validator, 'returns', False) is True:
+                    value = clean_value
             except ModelValidationError as exc:
                 errors.update(exc.messages)
             except StopValidation as exc:
--- a/schematics/validate.py
+++ b/schematics/validate.py
@@ -1,3 +1,5 @@
+import functools
+
 from .exceptions import BaseError, ValidationError, ModelConversionError
 from .transforms import import_loop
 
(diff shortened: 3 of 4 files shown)


lkraider:
What is the rationale here? I think this conflates validation and conversion with no gain.

bintoro:
@lkraider That's a good point. I wouldn't go so far as to say it conflates validation and conversion, but a poorly implemented validator might indeed cause trouble by returning a value of the wrong type.

In principle I would be fine with leaving it up to developers not to mess up the type, but since it's very easy to guard against this by routing the return value through `to_native()` when it has changed, I suppose why not.

As for the rationale, I haven't really questioned it because eveything implies it's intended to work this way, and the idea seems reasonable. For example, a `StringType` field `first_name` might strip trailing spaces from the value. Obviously, a name shouldn't contain trailing spaces, but it's not worth raising an error about.

Conceptually, such a sanitizing operation would seem to belong in the validation step as opposed to the conversion stage. Also, it makes sense that you can attach these kinds of normalization rules to individual fields as validators instead of having to make a whole new type every time.

lkraider:
In your example I would create a `UsernameType` to handle the cleanup on `to_native` conversion. 

As a counter-example to yours, imagine a case where your `first_name` is of `StringType(min_length=4)` and you add a custom validator functions that does the whitespace trim.

Now imagine it receives a string of four empty spaces `"    "`. Depending at which point your cleanup runs will affect the validation, either accepting the `min_length` or not, losing validation determinism. To fix we then have to be careful about chaining validators and then include some sort of priority ordering, which is prone to breakage and requires knowledge of the full chain.

The point about keeping conversion vs validation as different function signatures where one return a type and the other return a bool (just conceptually, in practice we preferred the exception handling for the extra context information) is a requirement to keep things simple and deterministic.

That's why I worked on pinning the core (at least on my fork, I have not validated 1.0 for myself yet), otherwise things fly out of control when users start implementing Types that affect the loop chain.

bintoro:
Okay, I'm convinced.

I would still like to avoid subclassing `StringType` in the example case, but I now agree that it makes all kinds of sense to separate such preprocessing from the validation step. Not only will it be a more robust approach, a normalization step would mainly be applicable to data import only, so it'd be redundant and maybe even harmful to include it in every validation run.

Building on the idea drafted in PR #269, I could imagine the following stages of import processing:

  1. *Unmarshalling*: manipulating the raw data to fit the models
  2. *Converting*: primitive to native
  3. *Normalizing*: manipulating the native value
  4. *Validating*

It would probably even suffice to just have a slot for one normalization function per field that would be called after (or as part of) `to_native()`.

In fact, the whole thing could be encapsulated in `BaseType` if `to_native()` was only defined there and subclasses had something like `_to_native()` that would be invoked by the base class.

lkraider:
Subclassing allows you to just call the `super().to_native` and do your extra cleanup afterwards.

We could add a chaining mechanism for import processing so that you could maybe write `StringType(normalize=str.strip)` for the simple cases, although I would still advocate subclassing as the better choice for code reuse and code maintenance, but I am not against it.

Issue #357, opened 2015-11-05
TimeStampType critically broken


bintoro:
I've just reviewed `TimeStampType`, and it seems to be performing quite interesting date arithmetic:

```python
def date_to_timestamp(cls, value):
    ...
    return (delta.days * 24 * 3600) + delta.seconds + delta.microseconds
```

So yeah, it's adding microseconds on top of seconds. :koala: This means that, if the datetime includes microseconds, the results are essentially random timestamps 0 to 999999 seconds (~12 days) into the future from the intended time.

Also, while `DateTimeType` creates timezone-naive datetimes, if such an object is passed to `TimeStampType.to_primitive()`, it assumes local time. This is not a very good assumption because the results then depend on where the application happens to be deployed. I say either require the `tzinfo` component or assume UTC when it's missing. Any sane application should store datetimes in UTC anyway.

My first instinct to fix this is by removing the `types.temporal` module altogether and putting a working `TimeStampType` directly in `types.base`. This way, developers using the class will hopefully be alerted to discover documentation about these issues.

I presume the separate `temporal` module exists because of the dependency on `dateutil`, but that's not needed for anything besides getting the local timezone, which we shouldn't even be doing.

bintoro:
Resolved by #358.

Issue #173, opened 2013-10-09
ISO8601 timezone datetime


tomasd:
According to http://en.wikipedia.org/wiki/ISO_8601 datetime could be of 3 types:
- local datetime
- utc datetime
- timezone offset datetime

Currently only the 1st option is available in schematics. Would it be possible to support  the other ones? 

For instance this call parses timezone correclty from ISO8601:
import dateutil.parser
yourdate = dateutil.parser.parse(datestring)

bintoro:
How about modifying `DateTimeType` so that it would accept a pluggable parser object? That way developers could use an external library of their choosing, be it dateutil, Arrow, or whatever.

bintoro:
Fixed via #358.

PR #358, opened 2015-11-11
DateTimeType & TimestampType major rewrite

bintoro:
Here's a PR that includes a completely new `TimestampType` (#357, #271) and a `DateTimeType` with a bunch of new features (see #173, #289).

  * The built-in parser accepts Unix timestamps and all ISO 8601 variants between `YYYY-MM-DD hh:mm` and `YYYY-MM-DDThh:mm:ss.ssssss±zz:zz`.

  ```python
  >>> s = '2015-11-08T12:34:56.0330+02:00'
  >>> DateTimeType().to_native(s)
  datetime.datetime(2015, 11, 8, 12, 34, 56, 33000, tzinfo=offset_timezone(+02:00))
  ```

  * It understands timezones and can be configured to either preserve the values or convert to UTC.

  ```python
  >>> DateTimeType(convert_tz=True).to_native(s)
  datetime.datetime(2015, 11, 8, 10, 34, 56, 33000, tzinfo=utc_timezone())
  ```

  * You can plug in external parsers such as dateutil and Arrow.

  ```python
  >>> import dateutil.parser
  >>> DateTimeType(parser=dateutil.parser.parse).to_native('Fri, 06 Nov 2015 09:18:42 -0700')
  datetime.datetime(2015, 11, 6, 9, 18, 42, tzinfo=tzoffset(None, -25200))
  ```

  ```python
  >>> import arrow.parser
  >>> arrow_parse = lambda value: arrow.parser.DateTimeParser('fr_fr').parse(value, 'Do MMMM YYYY HH:mm')
  >>> DateTimeType(parser=arrow_parse).to_native(u'1er août 2015 13:30')
  datetime.datetime(2015, 8, 1, 13, 30)
  ```

There are 3 settings dealing with timezone stuff:

  * **tzd**: Time zone policy
      * `'require'`: Values must specify a time zone.
      * `'allow' `: Values both with and without a time zone designator are allowed.
      * `'utc'    `: Like `'allow'`, but values with no time zone information are assumed UTC.
      * `'reject'`: Values must not specify a time zone. This also prohibits timestamps.

  * **convert_tz**
      * `True`:  Convert the datetime to UTC based on its time zone offset.
      * `False`: Don't convert. Keep the original time and offset intact.

  * **drop_tzinfo**
      When all values are known to be in UTC, can be set to `True` to discard the UTC `tzinfo` objects and make naive `datetime` instances instead.

### Compatibility

Since June, `DateTimeType` has accepted UTC times (designator "Z") as well as local times. It discards the timezone information, however, and only produces naive `datetime` objects. I'm concerned that interpreting unknown timezones as UTC might be too bold of an assumption.

It is possible to closely emulate the old behavior. Specifying `tzd='utc', convert_tz=True, drop_tzinfo=True` would coerce everything to UTC and produce naive datetimes. This is actually a very good setup in practice, but it's assuming a lot.

Another option is to revert the change regarding the Z designator and go back to only accepting timezoneless values by default. That way, users would have to deliberately activate time zone support, which means they would have to make a conscious choice about how to handle them.


bintoro:
With the latest commits I think this PR has become mergeable.

In short:

* `DateTimeType` gets time zone support plus a built-in parser.
* `TimestampType` becomes more usable and bug-free.
* The `types.temporal` module is **removed**.

No external dependencies are needed. The only open question is what default behavior to choose w.r.t. timezones.

PR #310 added support for a UTC designator in the parsing format but did nothing to actually handle the time zone information. Going forward, there are several reasonable options:

* (A) Accept all kinds of inputs and attach `tzinfo` only if the input contains time zone information.

  * `2015-01-01 12:00:00-07:00` would be imported as `datetime(2015, 1, 1, 12, 0, tzinfo=offset_timezone(-07:00))`
  * Pros: Has the least surprising `to_native` outcome.
  * Cons: Resulting objects may be either naive or aware depending on input.

* (B) Accept all kinds of inputs. Convert timezone-aware values to UTC and drop the `tzinfo` objects.

  * `2015-01-01 12:00:00-07:00` would be imported as `datetime.datetime(2015, 1, 1, 19, 0)`
  * Pros: An optimized setup in terms of actual application design. Compatible; preserves current behavior for both presently accepted formats.
  * Cons: Opinionated. Not viable for fields that must be able to accept both local times and timezone-aware values, but I believe this to be rare.

* (C) Revert #310 and stop accepting values with timezones by default.

  * `2015-01-01 12:00:00-07:00` would be rejected.
  * Pros: Having to explicitly enable time zone support means that there will be no surprises in the field's behavior.
  * Cons: Kinda 1988.

I don't have a definite favorite, but I'm leaning towards B because it represents a kind of best practice. In most cases it's ideal to work in UTC internally; in addition, dropping the redundant `tzinfo` components helps prevent bugs. While this setup would keep the current conflation between `<datetime>` and `<datetime>Z`, it's probably not a big deal, since local datetimes mainly appear in user-facing apps and aware datetimes in APIs.

j2labs:
I think I like B best. My thinking is that the data internal to a Model is
probably best kept in it's most raw form. Further, I have seen sooooo many
mistakes come from folks doing timezones incorrectly, so I like the idea of
nudging people towards getting it right.

On Fri, Nov 20, 2015 at 4:52 PM, Kalle Tuure <notifications@github.com>
wrote:

> With the latest commits I think this PR has become mergeable.
>
> In short:
>
>    - DateTimeType gets time zone support plus a built-in parser.
>    - TimestampType becomes more usable and bug-free.
>    - The types.temporal module is *removed*.
>
> No external dependencies are needed. The only open question is what
> default behavior to choose w.r.t. timezones.
>
> PR #310 <https://github.com/schematics/schematics/pull/310> added support
> for a UTC designator in the parsing format but did nothing to actually
> handle the time zone information. Going forward, there are several
> reasonable options:
>
>    -
>
>    (A) Accept all kinds of inputs and attach tzinfo only if the input
>    contains time zone information.
>    - 2015-01-01 12:00:00-07:00 would be imported as datetime(2015, 1, 1,
>       12, 0, tzinfo=offset_timezone(-07:00))
>       - Pros: Has the least surprising to_native outcome.
>       - Cons: Resulting objects may be either naive or aware depending on
>       input.
>    -
>
>    (B) Accept all kinds of inputs. Convert timezone-aware values to UTC
>    and drop the tzinfo objects.
>    - 2015-01-01 12:00:00-07:00 would be imported as datetime.datetime(2015,
>       1, 1, 19, 0)
>       - Pros: An optimized setup in terms of actual application design.
>       Compatible; preserves current behavior for both presently accepted formats.
>       - Cons: Opinionated. Not viable for fields that must be able to
>       accept both local times and timezone-aware values, but I believe this to be
>       rare.
>    -
>
>    (C) Revert #310 <https://github.com/schematics/schematics/pull/310>
>    and stop accepting values with timezones by default.
>    - 2015-01-01 12:00:00-07:00 would be rejected.
>       - Pros: Having to explicitly enable time zone support means that
>       there will be no surprises in the field's behavior.
>       - Cons: Kinda 1988.
>
> I don't have a definite favorite, but I'm leaning towards B because it
> represents a kind of best practice. In most cases it's ideal to work in UTC
> internally; in addition, dropping the redundant tzinfo components helps
> prevent bugs. While this setup would keep the current conflation between
> <datetime> and <datetime>Z, it's probably not a big deal, since local
> datetimes mainly appear in user-facing apps and aware datetimes in APIs.
>
> —
> <https://github.com/schematics/schematics/pull/358#issuecomment-158535433>
> .
>



-- 
jms dnns
<https://twitter.com/jmsdnns>

bintoro:
Yeah, exactly.

I'm just slightly uneasy about dropping the only piece of information indicating that the original input contained a tz offset, when it may or may not have been there.

A less controversial variant of plan B could involve converting tz-aware values to UTC but stopping short of dropping the `tzinfo` objects:

  * value with an offset → `datetime.datetime(<converted datetime>, tzinfo=UTC)`
  * value without an offset → `datetime.datetime(<raw datetime>, tzinfo=None)`

This would preserve the distinction between a local time *n* and UTC time *n*Z while offering some protection against stupid code that ignores the tzinfos.

Anyway, the fact of the matter is there's no one size fits all solution, since the ideal setup totally depends on the use case. I'll leave the PR open for a while more in case we get more input.

j2labs:
I think we should preserve the TZ info.

The user can easily go back to the original input this way.

On Mon, Nov 23, 2015 at 1:56 PM, Kalle Tuure <notifications@github.com>
wrote:

> Yeah, exactly.
>
> I'm just slightly uneasy about dropping the only piece of information
> indicating that the original input contained a tz offset, when it may or
> may not have been there.
>
> A less controversial variant of plan B could involve converting tz-aware
> values to UTC but stopping short of dropping the tzinfo objects:
>
>    - value with an offset → datetime.datetime(<converted datetime>,
>    tzinfo=UTC)
>    - value without an offset → datetime.datetime(<raw datetime>,
>    tzinfo=None)
>
> This would preserve the distinction between a local time *n* and UTC time
> *n*Z while offering some protection against stupid code that ignores the
> tzinfos.
>
> Anyway, the fact of the matter is there's no one size fits all solution,
> since the ideal setup totally depends on the use case. I'll leave the PR
> open for a while more in case we get more input.
>
> —
> <https://github.com/schematics/schematics/pull/358#issuecomment-159027616>
> .
>



-- 
jms dnns
<https://twitter.com/jmsdnns>

bintoro:
Ha. I was just about to post the same thing.

I've thought about this a lot since yesterday, and I keep coming to the conclusion that the current defaults (plan A, no conversion to UTC) is the right way to go.

Any other option involves losing some aspect of the raw data, and I don't think Schematics should do that. Also, this setup can work out of the box for every use case, and then users can go from there and make it more restrictive rather than the other way around.

j2labs:
I see. So Plan A is your thoughts on how to preserve it.

Well I can dig that.

The goal for schematics is to convert into a raw type (the native type) and
I don't have a strong opinion on whether or not that includes a timezone
conversion. It's an edge case. So I can lean either way as long as the
original value is easily attainable.

Plan A, however, has less of a performance hit, so I'm happy to lean that
way.

On Tue, Nov 24, 2015 at 10:14 AM, Kalle Tuure <notifications@github.com>
wrote:

> Ha. I was just about to post the same thing.
>
> I've thought about this a lot since yesterday, and I keep coming to the
> conclusion that the current defaults (plan A, no conversion to UTC) is the
> right way to go.
>
> Any other option involves losing some aspect of the raw data, and I don't
> think Schematics should do that. Also, this setup can work out of the box
> for every use case, and then users can go from there and make it more
> restrictive rather than the other way around.
>
> —
> <https://github.com/schematics/schematics/pull/358#issuecomment-159299457>
> .
>



-- 
jms dnns
<https://twitter.com/jmsdnns>

bintoro:
Yeah well that's the thing, in plan B the original input would *not* be attainable.

For instance an email timestamp `Tue, 24 Nov 2015 07:17:54 -0800` tells you the local wall clock time of the sender. If Schematics were to silently turn this into `2015-11-24 15:17:54+0000`, a part of the information would be lost, and that's kinda troubling. Only the application author can determine what's important, and that's why I'm thinking irreversible conversions should likely be opt-in.

j2labs:
What Plan A and Plan B represent to me is whether or not we want to convert
to UTC and treat UTC like a more normal form of a regular date.

If Plan B then represents conversion to UTC, we'd need to also track what
tz the date *was,* which might require an additional field that isn't
explicitly mentioned, such that it is part of the output for exports. I am
not keen on this particular part.

But, I do find the idea of treating dates as UTC by default appealing.
Perhaps, just not inside schematics.

On Tue, Nov 24, 2015 at 10:35 AM, Kalle Tuure <notifications@github.com>
wrote:

> Yeah well that's the thing, in plan B the original input would *not* be
> attainable.
>
> For instance an email timestamp Tue, 24 Nov 2015 07:17:54 -0800 tells you
> the local wall clock time of the sender. If Schematics were to silently
> turn this into 2015-11-24 15:17:54+0000, a part of the information would
> be lost, and that's kinda troubling. Only the application author can
> determine what's important, and that's why I'm thinking irreversible
> conversions should likely be opt-in.
>
> —
> <https://github.com/schematics/schematics/pull/358#issuecomment-159306198>
> .
>



-- 
jms dnns
<https://twitter.com/jmsdnns>

lkraider:
I can agree with A being default. B could be achieved with a UTCDateTimeType if someone wants.

j2labs:
YES. that seems so obvious now that you've said it... hahaha. that's
exactly the right way to do it

On Tue, Nov 24, 2015 at 10:50 AM, Paul Eipper <notifications@github.com>
wrote:

> I can agree with A being default. B could be achieved with a
> UTCDateTimeType if someone wants.
>
> —
> <https://github.com/schematics/schematics/pull/358#issuecomment-159310370>
> .
>



-- 
jms dnns
<https://twitter.com/jmsdnns>

bintoro:
Yeah, including derivative classes with some common configurations wouldn't hurt. It's just that there are so many scenarios that serious applications will likely want to set their own settings.

Want to always work in UTC? Okay, set `convert_tz=True`. But do you want to assume timezoneless values are in UTC too or disallow them? Then that's either `tzd='utc'` or `tzd='require'`. What about `tzinfo`? If the export format string includes the tzd, then you'll want it, otherwise maybe not...

But I guess we're in agreement now that plan A is the proper default because it preserves the original input. (It's also arguably the least surprising behavior as it follows the precedent set by the popular datetime libraries.)

bintoro:
Added a `UTCDateTimeType` that corresponds to plan B. Particularly when working with SQL databases *not* having tzinfos around is pretty clearly the safest choice.

Or should this be just a sample recipe in the docs instead of a bundled type?

j2labs:
Let's bundle it in. It's simple enough :)

On Tuesday, November 24, 2015, Kalle Tuure <notifications@github.com> wrote:

> Added a UTCDateTimeType that corresponds to plan B. Particularly when
> working with SQL databases *not* having tzinfos around is pretty clearly
> the safest choice.
>
> Or should this be just a sample recipe in the docs instead of a bundled
> type?
>
> —
> <https://github.com/schematics/schematics/pull/358#issuecomment-159368212>
> .
>


-- 
jms dnns
<https://twitter.com/jmsdnns>

Issue #316, opened 2015-07-10
strict=False doesn't propagate as expected


kstrauser:
When one model is nested inside another, importing that model doesn't pass `strict` through as expected. For example:

```python
    from schematics.models import Model
    from schematics.types import StringType
    from schematics.types.compound import ModelType

    class Inner(Model):
        foo = StringType()

    class Outer(Model):
        inner = ModelType(Inner)

    def main():
        d = {
            'inner': {
                'foo': 'foo value',
                'bar': 'bar value',
            },
        }

        outer = Outer(d, strict=False)
        print outer
        print outer.inner
```

This fails with:

```python
    Traceback (most recent call last):
      File "/private/tmp/schem/bin/main", line 9, in <module>
        load_entry_point('main==0.1.0', 'console_scripts', 'main')()
      File "/private/tmp/schem/schematics_rogue_field_issue/src/main/main.py", line 19, in main
        outer = Outer(d, strict=False)
      File "/private/tmp/schematics/schematics/models.py", line 240, in __init__
        self._data = self.convert(raw_data, strict=strict, mapping=deserialize_mapping)
      File "/private/tmp/schematics/schematics/models.py", line 287, in convert
        return convert(self.__class__, raw_data, **kw)
      File "/private/tmp/schematics/schematics/transforms.py", line 419, in convert
        partial=partial, strict=strict, mapping=mapping)
      File "/private/tmp/schematics/schematics/transforms.py", line 119, in import_loop
        raise ModelConversionError(errors, data)
    schematics.exceptions.ModelConversionError: {'inner': {'bar': 'Rogue field'}}
```

Instead, I would have expected it to successfully instantiate the `outer` object.

kracekumar:
```python
In [20]: class Inner(Model):
    foo = StringType()
   ....:

In [21]: class Outer(Model):
    inner = ModelType(Inner, strict=False)
   ....:

In [22]: d = {'inner': {
'foo': 'v',
'bar': 'bad'}}

In [23]: v = Outer(d, strict=False)

In [24]: v.inner
Out[24]: <Inner: Inner object>

In [25]: v.to_primitive()
Out[25]: {'inner': {'foo': u'v'}}

In [26]: v.inner.to_primitive()
Out[26]: {'foo': u'v'}
```

Right way to do is to pass `strict=False` during declaration. 

Setting at model declaration step allows granular control.

kstrauser:
[...]

PR #359, opened 2015-11-16
Propagate environment during recursive import/export/validation

bintoro:
A rewrite of #266.

This is a huge PR, but it's split into meaningful commits, each of which represents a working state with all tests passing.

`env` is now always passed to all `to_primitive`, `to_native`, and `export_loop` methods on types. Consequently, the standard `field_converter` interface also has `env` as a mandatory component: `field_converter(field, value, env)`.

The `env` object will be set up by either `import_loop` or `export_loop` during the first iteration if the application hasn't supplied it by then.

### `context`

The overloading of "context" with a number of uses has been resolved as follows:

  * `context` in `import_loop()` and `validate()` renamed to `trusted_data`
  * `context` as an argument to `to_native()`, `to_primitive()`, `validate()`, and `mock()` superseded by `env`
  * `context` as an argument to `expand` in `transforms.py` renamed to `expanded_data`
  * `context` when accessing the private context replaced by `env.context`.

Basically, `env.context` is just a recommendation for establishing a private namespace. Apparently the only thing in the library that accesses `env.context` is `MultiLingualString` that looks for a locale there.

### `export_loop` inspection gone

[DELETED]

### Miscellaneous changes

  * `ModelType.strict` option removed for now. To support something like this, we would need to differentiate between at least two flavors of settings:
    * static settings (specified as function defaults, model options, or `ModelType` definitions), where later (as in deeper level of recursion) overrides earlier
    * explicit settings (specified by application at runtime) that would override everything else.

  I suspect `ModelType.strict` may have been a band-aid to provide some degree of control where the runtime `strict` setting hasn't been doing anything.
  * `BaseType.validate_required()` removed. `import_loop` enforces `required=True` for fields directly on a model, so this was only needed for nested fields. They now get this validator in `MultiType.__init__()` as needed.

### Validators

`env` is now passed to all validators, including model-level validators. Although the object is absolutely needed only in compound type validators, it's probably best to make the change everywhere at once, since appending `env` to the argument list is all that is needed to adapt existing code.

### DIFF
--- a/schematics/transforms.py
+++ b/schematics/transforms.py
@@ -39,7 +39,7 @@ ExportContext = get_context_factory('ExportContext',
 
 
 def import_loop(cls, instance_or_dict, field_converter, trusted_data=None,
-                partial=False, strict=False, mapping=None, app_data=None, env=None):
+                partial=False, strict=False, mapping=None, app_data=None, context=None):
     """
     The import loop is designed to take untrusted data and convert it into the
     native types, as described in ``cls``.  It does this by calling
@@ -51,7 +51,7 @@ def import_loop(cls, instance_or_dict, field_converter, trusted_data=None,
         The class for the model.
     :param instance_or_dict:
         A dict of data to be converted into types according to ``cls``.
-    :param field_convert:
+    :param field_converter:
         This function is applied to every field found in ``instance_or_dict``.
     :param trusted_data:
         A ``dict``-like structure that may contain already validated data.
@@ -61,13 +61,20 @@ def import_loop(cls, instance_or_dict, field_converter, trusted_data=None,
         definitions. Default: False
     :param strict:
         Complain about unrecognized keys. Default: False
+    :param app_data:
+        An arbitrary container for application-specific data that needs to
+        be available during the conversion.
+    :param context:
+        An ``ImportContext`` object that encapsulates configuration options and
+        ``app_data``. The context object is created upon the initial invocation
+        of ``import_loop`` and is then propagated through the entire process.
     """
     if not isinstance(instance_or_dict, (cls, dict)):
         raise ModelConversionError('Model conversion requires a model or dict')
 
     mapping = mapping or {}
     app_data = app_data if app_data is not None else {}
-    env = env or ImportContext(partial, strict, mapping, app_data)
+    context = context or ImportContext(partial, strict, mapping, app_data)
 
     data = dict(trusted_data) if trusted_data else {}
     errors = {}
@@ -78,18 +85,18 @@ def import_loop(cls, instance_or_dict, field_converter, trusted_data=None,
             all_fields.add(field.serialized_name)
         if field.deserialize_from:
             all_fields.update(set(_listify(field.deserialize_from)))
-        if field_name in env.mapping:
-            all_fields.update(set(_listify(env.mapping[field_name])))
+        if field_name in context.mapping:
+            all_fields.update(set(_listify(context.mapping[field_name])))
 
     # Check for rogues if strict is set
     rogue_fields = set(instance_or_dict) - all_fields
-    if env.strict and len(rogue_fields) > 0:
+    if context.strict and len(rogue_fields) > 0:
         for field in rogue_fields:
             errors[field] = 'Rogue field'
 
     for field_name, field in iteritems(cls._fields):
         trial_keys = _listify(field.deserialize_from)
-        trial_keys.extend(_listify(env.mapping.get(field_name, [])))
+        trial_keys.extend(_listify(context.mapping.get(field_name, [])))
         if field.serialized_name:
             serialized_field_name = field.serialized_name
             trial_keys.extend((serialized_field_name, field_name))
@@ -108,13 +115,13 @@ def import_loop(cls, instance_or_dict, field_converter, trusted_data=None,
 
         try:
             if raw_value is None:
-                if field.required and not env.partial:
+                if field.required and not context.partial:
                     errors[serialized_field_name] = [field.messages['required']]
             else:
                 field_params = {
-                    'mapping': env.mapping.get('model_mapping', {}).get(field_name, None)
+                    'mapping': context.mapping.get('model_mapping', {}).get(field_name, None)
                 }
-                raw_value = field_converter(field, raw_value, env._branch(**field_params))
+                raw_value = field_converter(field, raw_value, context._branch(**field_params))
 
             data[field_name] = raw_value
 
@@ -130,7 +137,7 @@ def import_loop(cls, instance_or_dict, field_converter, trusted_data=None,
 
 
 def export_loop(cls, instance_or_dict, field_converter, role=None, 
-                raise_error_on_role=False, print_none=False, app_data=None, env=None):
+                raise_error_on_role=False, print_none=False, app_data=None, context=None):
     """
     The export_loop function is intended to be a general loop definition that
     can be used for any form of data shaping, such as application of roles or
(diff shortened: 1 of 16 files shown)


schematics/transforms.py
```diff
@@ -56,11 +56,17 @@ def import_loop(cls, instance_or_dict, field_converter, context=None,
...
-    if mapping is None:
-        mapping = {}
-    data = dict(context) if context is not None else {}
-    errors = {}
 
+    if not env:
+        env = ConfigObject()
+    elif not isinstance(env, ConfigObject):
+        env = ConfigObject(env)
+    env._setdefault('partial', partial)
```
edc: I think it makes more sense to have values passed as arguments overwrite those in `env`.

Having two ways to specify options is confusing. But I understand you don't want to break backward compatibility.

bintoro: Hmm. There may be some misunderstanding here. For the most part, the options are supposed to be specified just like before:

```python
foo.import_data({...}, strict=True, partial=True)
```

This PR does not intend to change that.

An application *may* also pass an `env` object, but there are only two reasons to do so:
  * you want to put some proprietary data in it
  * your custom type or `Model` class wants to change the parameters *in the middle of* the recursive import/export process and propagate the new settings from there on.

What this code does is it sets up the `env` container on the first invocation, and then the options will be frozen thereafter. Letting arguments overwrite the prior state on every call is precisely the problem this PR aims to fix.

Here's an example of how a call stack might look like before:

  1. `Model.__init__({...}, strict=True)`
  2. `Model.convert({...}, strict=True)`
  3. `convert({...}, strict=True)`
  4. `import_loop({...}, strict=True)`
  5. `ModelType.to_native(...)`
  6. `Model.import_data(...)`
  7. `Model.convert(...)`
  8. `convert(...)`
  9. `import_loop(...)`

Okay, back at `import_loop`. What's the value of `strict`? It's `False`, set by `convert()` at step 7. But obviously it should be `True`, since that's what was originally specified.

There were two ways to fix this:

  * make sure every function throughout the chain propagates every possible parameter as an argument, or
  * encapsulate the settings in a single object in the beginning and pass that around instead.

I opted for the latter approach because it's way cleaner and you can now add new options without touching everything.

If there's something that was easy to do before but becomes awkward due to this change, I'm sure we can address it somehow.

bintoro:
Okay, I think I get it now about `to_native()`. You might want to export a structure that can be fed to another library that understands standard types like `datetime.datetime` but does not understand models.

Fair enough, but in that case "native" is not a well-defined concept. Alternatively, `Model.to_native()` is a misnomer, since the end result is a mixed native/primitive representation.

Resolving what "native" means is out of scope for this PR, so I'm going to shelve the fifth commit.

lkraider:
I like the cleanup idea proposed and the patch as is seems good.

Do you have an idea how this could affect execution performance? What's the tradeoff in overhead/ease of use of passing ConfigObjects instead of dicts or maybe namedtuples? (this is just a question, don't block this patch acceptance based on it - I think performance should be a next goal for schematics as a whole).

bintoro:
Great idea about `namedtuple`. When I wrote #266, I assumed the container should be mutable, but I don't think that's actually the case at all, particularly now that we're designating `env.context` for arbitrary data.

`namedtuple` would be faster, yes, but it would also prevent bugs arising from accidental modification of the settings.

(Whenever `env` is modified locally, such as in `import_loop` where `mapping` is replaced by its subtree, the object *must* be copied — otherwise the changes would apply in parent scopes too. A `namedtuple` would enforce this.)


However, using `namedtuple` means that passing a partial `env` as an argument won't be so practical. So, ideally, `context` should get its own argument after all, and that brings us back to question of whether it's too outrageous to change the meaning of `context` in `import_loop` or if it should be renamed.

In general I totally agree about the performance concerns. It would be great if someone put together a testcase for trying out some optimization ideas.

lkraider:
I see no problem in re-purposing the `context` keyword if we are breaking compatibility, it was de-facto an internal API anyway (the new `trusted_data`). But calling it `user_data` might be clearer.

bintoro:
Now using `namedtuple` with a modified constructor that allows omitting values to make testing easier:

```python
>>> from schematics.transforms import ExportContext
>>> ExportContext(role='public')
ExportContext(role='public', raise_error_on_role=None, print_none=None, app_data=None)
```

bintoro:
Although the diff is all over the place, this is a rather straightforward PR, so I'm preparing to merge.

Last call regarding nomenclature:

**(1)** After @lkraider alleviated my concerns about reusing the retired `context` parameter, I'm now considering just using that instead of `env` because:

  * `env` can sound like an application-wide or library-wide thing, while `context` has all the right connotations

  * it would reduce the need for mechanical adaptation of old code, as the old `context` in most places is simply about to be replaced with the new `env`

  * `context` is no longer used in any conflicting meaning.

**(2)** The old `context` data bag is now called `app_data`, which I felt was more neutral and unambiguous than either `private_data` or `user_data`, but I'm open to being overruled on this too.

Thoughts?

***

If we make the switch from `env` to `context`, migrating custom types would go like this:

  * Add `context` as a third parameter to validator methods.
  * No need to touch method signatures that already include `context`.
  * In actual method code, replace `context` with `context.app_data` if you're using the data for something.

lkraider:
Can you list the flow of options/context now? For example, how will something like `validate(partial=True, context={partial=False})` behave? (I suppose the context will be overridden by the parameter?)

bintoro:
> For example, how will something like `validate(partial=True, context={partial=False})` behave? (I suppose the context will be overridden by the parameter?)

Taken literally, that form (a user-provided dict as `context`) no longer exists, since with the addition of the `app_data` parameter everything can now be specified as regular arguments. In fact, the `context` parameter should be removed altogether from the likes of `Model.validate()` that make up the main API.

Other than that, `context` always overrides any individual arguments, since the arguments could just be random defaults.

This now makes things very simple: up until the first invocation of `import_loop` or `export_loop` there is no context object — everything's in the parameters. Once we get to the loop function, it creates the context object if there was none, and from there on only `context` matters.

The only case where a context object needs to be created from scratch is when writing tests for a method that needs one:

  ```python
  mls = MultilingualStringType(default_locale='en_US')

  assert mls.to_primitive(
      {'en_US': 'snake', 'fr_FR': 'serpent'},
      context=ExportContext(app_data={'locale': 'fr_FR'})) == 'serpent'
              ^^^^^^^^^^^^^
  ```

lkraider:
Thanks for clearing it up, sounds great for me.

bintoro:
I believe this PR is now mature enough to be merged.

It resolves #316 and facilitates the addition of new conversion options in the future without causing disruption to existing methods.

chekunkov:
@bintoro hi, when is this change going to be released on PyPI?

bintoro:
@chekunkov I'm not sure. The patch is technically backward-incompatible, since all `to_*` methods and custom validators are now required to accept the `context` parameter.

If we want to be really strict about semantic versioning, this would have to wait until v2.0-alpha. If we accept mildly breaking changes in a minor release, this could be shipped pretty quickly in v1.2.0. I haven't decided.

bintoro:
Just FYI, I've decided to include this patch in v1.2.0 that will be released shortly.

I'm going to alleviate upgrading pains by making the `context` parameter optional on validator functions. If a validator doesn't declare the parameter, the function will be placed in a wrapper that drops the extra argument at runtime.

jacobstr:
I'm curious what the state of the 1.2 release is?

ryanhiebert:
I know that I'm using a 2.0 pre-release personally. You might wish to try it yourself, to see if this feature is in that, and if it works for your cases.

Issue #338, opened 2015-10-19
IntType will validate a string


jmsdnns:
```python
>>> from schematics.types import IntType
>>> IntType().validate('foo')
>>> 
```

This doesn't make sense.

jmsdnns:
Fixed with this [PR](https://github.com/schematics/schematics/pull/339)

bintoro:
Hmm. Let's talk about this.

If this was a bug, it wasn't the fault of `IntType`, so the patch only fixed a very special case.

By design, validator functions are supposed to be able to rely on the value *already being of the correct type*. This is because model validation always runs `field.to_native()` before `validate()`.

In other words, validating *a model* with an `IntType` field containing `"foo"` would have surely raised an error.

If you run `field.validate()` directly, you can get all kinds of weird results:

```python
>>> IntType(min_value=100).validate("1")
>>>
```

Basically, validators are simply not meant to be called directly unless you make sure the type is right.

The way to fix this is to take the `to_native()` call out of the field converter function defined in `validate.py` and put it in `BaseType.validate()` where it will be called regardless of who's invoking the validation. (Note that even this doesn't help if you call a specific validator such as `validate_choices()` directly.)

Can anyone think of a reason why `BaseType.validate()` should *not* call `to_native()` first?

PR #363, opened 2015-11-21
Make sure field.validate() uses a native type

bintoro:
Validator functions are supposed to operate on native values, so
it makes sense to perform the conversion in BaseType.validate()
as opposed to the validation loop. It is now possible to invoke
`field.validate()` directly, which is especially useful for testing.

***

See #338.

Reverses the patch introduced in #339, as it's now unnecessary, but keeps the test.

### DIFF
--- a/schematics/contrib/mongo.py
+++ b/schematics/contrib/mongo.py
@@ -41,10 +41,3 @@ class ObjectIdType(BaseType):
     def to_primitive(self, value, context=None):
         return str(value)
 
-    def validate_id(self, value):
-        if not isinstance(value, bson.objectid.ObjectId):
-            try:
-                value = bson.objectid.ObjectId(unicode(value))
-            except Exception:
-                raise ValidationError('Invalid ObjectId')
-        return True
--- a/schematics/models.py
+++ b/schematics/models.py
@@ -235,7 +235,7 @@ class Model(object):
         self._data = self.convert(raw_data, strict=strict, partial=partial,
                                   mapping=deserialize_mapping, app_data=app_data, context=context)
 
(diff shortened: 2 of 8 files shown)


bintoro:
A downside of this patch is that it would become more difficult to utilize a custom validation loop that skips the `to_native()` part. This is something you might want to do when performance is important and there's no danger of having incorrect datatypes on the model instances.

An argument in favor of the patch is that it makes the behavior of `Model.validate()` and `field.validate()` consistent in the sense that in both cases `field.to_native()` would be automatically run for you.

To accommodate both goals, `BaseType.validate()` could take an argument `convert=True/False` indicating whether `to_native()` should be performed there or not.

bintoro:
The final revision of this patch adds a `convert=True/False` switch to `field.validate()`, `Model.validate()`, and the main `validate()` functions. It's now possible to skip the `to_native` step that is normally bundled with validation.

This allows us to move the conversion task inside `field.validate()` without any loss of flexibility. As explained earlier, moving the `to_native()` call is the only correct way to "fix" #338.

The `convert` option turns out to be useful way beyond the original purpose of this PR. Consider:

```python
m = SomeModel(<raw data>) # primitive-to-native conversion
m.validate()              # implies unnecessary native-to-native conversion
```

Now you can say:

```python
m.validate(convert=False)
```

when the datatypes are known to be good and only need to be validated. This is great for performance when only trusted operations have been performed after the initial import.

There were at least two validators that essentially duplicated the functionality of `to_native`:
  * `NumberType.validate_is_a_number()`
  * `ObjectIdType.validate_id()`.

These methods are removed.

PR #364, opened 2015-11-21
Fix NumberType.to_native()

bintoro:
This I think is a bona fide bug:

```python
>>> IntType().to_native(2.999)
2
```

That can't be by design, right?

As an aside, what do we need `LongType` for? Even in Python 2.6 integers become longs as needed:

```python
>>> int(987358972389723895234234234897)
987358972389723895234234234897L
```

### DIFF
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -3,6 +3,7 @@ import re
 import datetime
 import decimal
 import itertools
+import numbers
 import functools
 import random
 import string
@@ -459,12 +460,12 @@ class NumberType(BaseType):
     }
 
     def __init__(self, number_class, number_type,
-                 min_value=None, max_value=None, coerce=False, **kwargs):
+                 min_value=None, max_value=None, strict=False, **kwargs):
         self.number_class = number_class
         self.number_type = number_type
         self.min_value = min_value
         self.max_value = max_value
-        self.coerce = coerce
+        self.strict = strict
 
         super(NumberType, self).__init__(**kwargs)
 
@@ -479,10 +480,11 @@ class NumberType(BaseType):
         except (TypeError, ValueError):
             pass
         else:
-            if (self.number_class is float  # Float conversion is strict enough.
-              or native_value == value      # Match numeric types.
-              or str(native_value) == value # Match numeric strings.
-              or self.coerce):
+            if self.number_class is float:  # Float conversion is strict enough.
+                return native_value
+            if not self.strict and native_value == value: # Match numeric types.
+                return native_value
+            if isinstance(value, (basestring, numbers.Integral)):
                 return native_value
 
         raise ConversionError(self.messages['number_coerce']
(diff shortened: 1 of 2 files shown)


bintoro:
Actually, I think the proper way to solve this is to introduce a `coerce=False/True` switch. There are a lot of objects that implement the `__int__()` method. Consider:

```python
>>>  IntType().to_native(uuid.uuid4())
295665622293270110940543243011398678078L
```

I suppose this might be desirable in some cases, but I don't think it's a good default to get an integer at any cost. The float truncation phenomenon is especially disturbing.

In a similar vein, `ListType` currently coerces an integer or a string into a single-element list. IMO that should be opt-in as well.

bintoro:
Okay, I've edited the patch to apply a stricter conversion by default but provide an option to revert to the old behavior.

Input                  |   New result       |   Old result
---------------------- | ------------------ | ----------------
3.0                    |   3                |   3      
3.2                    |   ERROR            |   3  
"3"                    |   3                |   3      
"3.0"                  |   ERROR            |   ERROR  
Decimal("3.0")         |   3                |   3      
Decimal("3.2")         |   ERROR            |   3
Fraction(30, 10)       |   3                |   3      
Fraction(7, 2)         |   ERROR            |   3  
uuid.uuid4()           |   ERROR            |   long(...)

bintoro:
Before merging this I'd like to solicit opinions on whether the patch is going far enough. Personally I think not, because accepting even `3.0` (float) as an integer is pushing it.

Perhaps, instead of having a switch to enable the outrageous old behavior (`3.2 => 3`), we could make the default mode a bit stricter than today, and then have a switch for enforcing *even stricter* standards:

Input                  |   Strict mode      |   New normal mode
---------------------- | ------------------ | ------------------
3                      |   3                |   3      
3.0                    |   ERROR            |   3      
3.2                    |   ERROR            |   ERROR
"3"                    |   3                |   3      
"3.0"                  |   ERROR            |   ERROR
Decimal("3.0")         |   ERROR            |   3      
Decimal("3.2")         |   ERROR            |   ERROR
Fraction(30, 10)       |   ERROR            |   3      
Fraction(7, 2)         |   ERROR            |   ERROR  
uuid.uuid4()           |   ERROR            |   ERROR

That is, in the new normal:
* No rounding or destructive truncation, but floats etc. are okay if they equal the converted integer.
* No more accepting types that implement `__int__()` but don't want to be compared as ints (e.g., `uuid.UUID`). I think if someone wants to store an obviously non-numeric type in an `IntType` field, it should come as no surprise that they first have to call `int()` on it.

This would be mostly compatible with the current situation, and if not, it's probably a bug in the application anyway. (Why would you knowingly store 4.9 in an `IntType` field in the anticipation that it will becoma 4?)

In the strict mode, acceptable values would be only
* actual integers (`isinstance(value, numbers.Integral) == True`)
* strings representing an integer.

bintoro:
I believe "strict or stricter" is a more useful menu of configurations than "strict or silly". The interim commit remains, in case someone wants to explore the silly behavior again in the future.

PR #365, opened 2015-11-22
Correctly decode bytestrings in Python 3

bintoro:
Fixes and simplifies Unicode routines. Also adds tests which were
notably absent.

***

##### Before patch

Python 2:

```python
>>> StringType()('\xc3\xa4').encode('utf-8')
'\xc3\xa4'
```

Python 3:

```python
>>> StringType()(b'\xc3\xa4').encode('utf-8')
  ...
schematics.exceptions.ConversionError: ["Couldn't interpret 'b'\\xc3\\xa4'' as string."]
```

### DIFF
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -21,23 +21,11 @@ except ImportError:
 
 try:
     basestring #PY2
+    bytes = str
 except NameError:
     basestring = str #PY3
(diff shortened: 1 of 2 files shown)


bintoro:
This seems like an uncontroversial bugfix, so merging.

Issue #366, opened 2015-11-26
Allow serializing only imported fields with to_primitive and to_native


tuukkamustonen:
As in the subject: allow something like:

    class Song(Model):
        ...
        class Options:
            serialize_only_dirty=True

Use case: I'm validating PATCH requests so wrapping the request payload into schematics model instance and would like to pass the instance deeper into code and serialize it later on when needed.

The problem is, my API allows clearing fields by setting them to `null`. So, by default all the (even non-required) fields are converted and thus cleared and I cannot use `serialize_when_none`, because then nulls wouldn't get through at all.

I could pass my request payload as is, yeah, but I'd rather pass schematics model and suffer the performance hit when needlessly serializing model into dict again...

bintoro:
This is in the works.

As of now, Schematics offers no way to distinguish an explicit `None` from "not set" (or "missing"). This causes all kinds of headaches when dealing with partial data, as you've discovered.

I've wanted to fix this for a long time, but work on the feature was halted due to lack of feedback and other issues. But it's coming in the next few weeks.

Right now, I'm planning to deprecate `serialize_when_none` and replace it with a setting like `export_level`, which might look something like this:

  * 0: never export
  * 1: don't export when empty (`None`, `[]`, `{}`, undefined)
  * 2: don't export when `None` or undefined
  * 3: don't export when undefined
  * 4: always export

Levels 1 and 4 are what's currently being offered via `serialize_when_none`.

Also see #360, which deals with applying defaults to incoming data.

tuukkamustonen:
That approach and those levels sound very reasonable, and would fulfill my needs.

I don't know schematics internals, but would it make sense to allow providing a custom serializer (per model or per field)? And then bundle implementation for the ones you listed.

Also #360 makes sense.

Should I close this ticket - you said it's in the words, is there another ticket somewhere?

bintoro:
> I don't know schematics internals, but would it make sense to allow providing a custom serializer (per model or per field)? And then bundle implementation for the ones you listed.

Incidentally I'm working on this, too.

The serialization mechanism is already extensible up to a point, but it could be better. Specifically, I want to generalize the export system so that, instead of choosing between `to_native()` and `to_primitive()`, you could specify a mixed representation as well.

Consider:

```python
class Foo(object):
    def __init__(self, x, y):
        self.x, self.y = x, y

class FooType(BaseType):
    def to_native(self, value, context):
        # convert dict to Foo instance
    def to_primitive(self, value, context):
[...]

Issue #337, opened 2015-10-15
export_loop() vs. to_primitive()


bintoro:
All the compound types define both `export_loop` and `to_primitive` methods, but `to_primitive` never seems to be called.

Am I right in assuming that, for compound types, `to_primitive` is just a historical remnant that's been superseded by `export_loop`? And that there are two differently named methods simply because the argument lists are different?

If yes, then obviously it would be ideal to unify the methods so that `field.to_primitive(...)` would accomplish the same thing for both simple and complex types. There's going to be an opportunity to do just that when #266/#316/#318 is implemented, since it involves a one-time method signature change across the board in order to future-proof against further such changes.

jmsdnns:
I believe this is correct. The original idea was that you could always call `to_primitive` on a type and it would just work. But `export_loop` became the way we'd pass around a transformation function, instead of assuming anything about whether or not a primitive was desired.

From then on, the compound types were containers and thus pathways towards getting to types that actually had `to_primitive` or `to_native` on them.

Would love a PR removing those methods, assuming tests are still good.

bintoro:
The redundant methods have been axed via #342. Not having them around helps avoid confusion, but I'll leave this issue open, as the ultimate solution is a single interface that works for both simple and compound types.

jmsdnns:
I'm not sure I agree on the single interface. I think an export loop is only useful when you are unpacking something with multiple values in it, like a model or compound type.

But, I think what you're talking about gets us closer to thinking about how to remove `ModelType` and possibly just have `Model` instances behave more like types. It feels to me like Models should just behave like types. They already have a `to_primitive` and `to_native` implementation.

Maybe it's just that I haven't been active in this codebase in a few years, but the `ModelType` really seems useless to me today. [Look at this](https://github.com/schematics/schematics/blob/development/schematics/types/compound.py#L59), it doesn't do anything the model wouldn't already do.

Hmm.

bintoro:
Passing around the transformation function will not be a problem. See #266.

What I want is to get rid of kludgy conditional paths like [this](https://github.com/schematics/schematics/blob/development/schematics/transforms.py#L173) and [this](https://github.com/schematics/schematics/blob/development/schematics/types/serializable.py#L49).

jmsdnns:
Hmm. I guess if these details are tucked away behind `to_native` and
`to_primitive` (_which likely still call export_loop_), we can always call
field_converter and know it'll work.

[...]

Issue #324, opened 2015-07-19
What would you incompatibly break?


kstrauser:
We've learned a lot along the way. Knowing what you know now, if you were implementing Schematics from scratch, what would you do differently? I'm *not* suggest we do this - I was just discussing this with @jvantuyl and it was an interesting exploration.

I'll lead:

- I'd toss Python 2.6 compatibility. I don't think there's much demand for it anymore and it's officially retired.
- Implement @bintoro's PR #266 metadata patch or something very similar.

Your turn. What would you personally do differently if this were a brand new project?

jvantuyl:
First and foremost, I don't want to give the impression that schematics isn't a wonderful piece of engineering.  I use it every day and it does what it was designed to do... quite well.

That said, Python 3 has brought with it quite a few changes.  Some of the data I've encountered legitimately has cycles.  Other data has pretty unusual requirements around correctness (i.e. a dict with a key that could be an int, a string, or another dict with a certain schema).  Finally, there are some dark corners of the codebase where we "do the right thing" that can sometimes manifest as "behavior" (Random references:  #53 + #54 + #76 [semi-intentional sloppy coercion across all of our types], #167 + #184 + #185 + #186 [polymorphic data structures are kind of a pain], #194 [bytes makes this complicated again], #210 [no easy way to validate a whole model without piggy-backing on a field or type validator], #266 + #316 + #318 [issues right at the heart of composition / strictness]).

There's a lot to be done that can adapt to the above issues.  Off the top of my head:

* Make coercion more controllable (i.e. right now, our models have optional strict interpretation but our fields are always loose)
* Make StringFields encoding-smart (i.e. encoding=None [which passes through a string] versus encoding=utf_8/ascii/utf_16/shift_jis/anything_in_codecs/rot13_lol).
* Make ByteFields for when we actually want untouched byte strings.
* Perhaps a relation API to sew together different datasets that have references.
* Find a way to solidly address the concerns of polymorphic data (this is easier when encoding than when decoding)
* Global validators on models.
* Less magic.  We're have a lot of complexity for what we get out of things like meta-classes, I think.
* Perhaps trade our transform loops for some recursion?  I can't help but think that the polymorphism stuff would be easier if we could delegate to the models for some of the composition.

There's probably plenty more, but the overall difference I'm shooting for is:

* take a step back and see if we can refactor the core to better tackle the things it makes difficult
* fit better into Python 3 (the bytes / strings thing is new, very touchy, and encoding is our concern)
[...]

PR #368, opened 2015-11-28
Unify import/export interface across all types

bintoro:
This is based on the ideas I talked about [here](https://github.com/schematics/schematics/issues/337#issuecomment-157871276) and [here](https://github.com/schematics/schematics/issues/324#issuecomment-155924477). The patch standardizes on methods called `convert` and `export`:

```python
def convert(self, value, context):
    # import and convert to native
```
```python
def export(self, value, format, context):
    # export according to format
```

where `format` is either `PRIMITIVE` or `NATIVE` (or a custom format; a type could choose to support more than these two).

Compound types implement these methods directly, while basic types can just map them to `to_native` and `to_primitive` as appropriate.

The bundled compound types no longer have any concept of primitive vs. native, as it makes no difference to them.

### Field converters

The overhaul gets rid of the famous `hasattr(field, 'export_loop')` conditionals. Now `export_loop` can just apply `field_converter` to every field and not care about what it does.

The fact that everything now passes through the field converter function opens up all kinds of possibilities for customizing the export process. To make this capability more prominent, `transforms.py` now defines the field converter interface:

```python
class FieldConverter(object):
    def __call__(self, field, value, context):
        raise NotImplementedError
```

The module includes a basic implementation that also supports mixed representations as described below.

### Mixed formats

In [this comment](https://github.com/schematics/schematics/issues/366#issuecomment-159982673) I provided a rationale for mixed representations, where some fields would be output as native and others as primitives. My initial plan was to introduce an "export profile" object that would have encapsulated this information.

However, since ultimately the field converter is in charge of the code that would query the export profile, it made sense to simply assign this responsibility to the field converter and not specify the details.

The bundled `ExportConverter` class supports configurations like "export everything as native except types *x*, *y*, and *z* as primitive" and vice versa.

### Compatibility

Basic types continue to work without modifications. The mapping from `field.export()` and `field.convert()` to the actual conversion methods is taken care of by `BaseType` unless overridden.

`Model.to_native()` used to return a mixed representation where models were converted to dictionaries but everything else remained in its native form. This behavior is now available as `Model.to_dict()`. The new `Model.to_native()` runs the export loop and applies roles but returns an instance, not a `dict`.

### DIFF
--- a/schematics/transforms.py
+++ b/schematics/transforms.py
@@ -5,8 +5,10 @@ import itertools
 
 from six import iteritems
 
-from .datastructures import OrderedDict, get_context_factory
+from .common import NATIVE, PRIMITIVE, EMPTY_LIST, EMPTY_DICT
+from .datastructures import OrderedDict, get_context_factory, DataObject
 from .exceptions import ConversionError, ModelConversionError, ValidationError
+from .types.compound import ModelType
 
 try:
     basestring #PY2
@@ -32,13 +34,13 @@ except:
 ###
 
 ImportContext = get_context_factory('ImportContext',
-                    'partial, strict, mapping, app_data')
+                    'field_converter, partial, strict, mapping, app_data')
 
 ExportContext = get_context_factory('ExportContext',
-                    'role, raise_error_on_role, print_none, app_data')
+                    'field_converter, role, raise_error_on_role, print_none, app_data')
 
 
-def import_loop(cls, instance_or_dict, field_converter, trusted_data=None,
+def import_loop(cls, instance_or_dict, field_converter=None, trusted_data=None,
                 partial=False, strict=False, mapping=None, app_data=None, context=None):
     """
     The import loop is designed to take untrusted data and convert it into the
@@ -74,7 +76,7 @@ def import_loop(cls, instance_or_dict, field_converter, trusted_data=None,
 
     mapping = mapping or {}
     app_data = app_data if app_data is not None else {}
-    context = context or ImportContext(partial, strict, mapping, app_data)
+    context = context or ImportContext(field_converter, partial, strict, mapping, app_data)
 
     data = dict(trusted_data) if trusted_data else {}
     errors = {}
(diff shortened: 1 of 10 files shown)


bintoro:
I'm gonna go ahead and merge, since any sort of uniform interface is a clear improvement to the status quo, and the undefined branch already builds on this.

If there are any controversial details, there's plenty of time to tweak those before this ends up in a stable release.

schematics/transforms.py
```diff
@@ -426,54 +424,48 @@ def blacklist(*field_list):
 ###
 
 
-def convert(cls, instance_or_dict, trusted_data=None, partial=True, strict=False,
-            mapping=None, app_data=None, context=None):
-    field_converter = lambda field, value, context: field.to_native(value, context)
-    data = import_loop(cls, instance_or_dict, field_converter, trusted_data=trusted_data,
-                       partial=partial, strict=strict, mapping=mapping, app_data=app_data, context=context)
-    return data
+class FieldConverter(object):
```
lkraider: Why the superclass/inheritance?

schematics/transforms.py
```diff
@@ -426,54 +424,48 @@ def blacklist(*field_list):
...
+    def __call__(self, field, value, context):
+        raise NotImplementedError
 
-def to_native(cls, instance_or_dict, role=None, raise_error_on_role=True, app_data=None, context=None):
-    field_converter = lambda field, value, context: field.to_native(value, context)
-    data = export_loop(cls, instance_or_dict, field_converter, role=role,
-                       raise_error_on_role=raise_error_on_role, app_data=app_data, context=context)
-    return data
 
+class ExportConverter(FieldConverter):
```
lkraider: Can't we replace this class with function currying?

schematics/transforms.py
bintoro: The big idea here is to define an extensible interface. Instead of passing a simple converter function you can pass a callable that may also implement pre- and post-conversion hooks (which allow the application to examine and mutate the dataset as a whole).

For example, I've written a pluggable converter class that uses this facility to export [JSON API](http://jsonapi.org/) -compliant payloads instead of flat dictionaries.

The default implementation is super simple and could indeed be replaced with a plain function. It mainly exists to serve as a model implementation.

(Note that the hooks were implemented later — have a look at the current revision.)

schematics/transforms.py
bintoro: I can definitely agree with simplifying the built-in converters. I suppose it could make the program flow easier to understand. We can have a separate recipe to illustrate an advanced use case.

schematics/transforms.py
lkraider: Makes sense if we have the extra machinery in the converter API for processing the data. Do we have a definition on where/when these pre- / post- hooks are called?

PR #372, opened 2015-12-05
Add support for undefined values

bintoro:
### Basics

As before, objects are populated by default.

```python
class M(Model):
    intfield = IntType()
    stringfield = StringType()
    defaultfield = StringType(default='foo')
```

```python
>>> m = M()
>>> m.to_primitive()
{'intfield': None, 'stringfield': None, 'defaultfield': 'foo'}
```

The `del` statement, which [used to be completely broken](https://github.com/schematics/schematics/pull/349), can now be used to clear an individual value.

```python
>>> del m.intfield
>>> m.to_primitive()
{'defaultfield': 'foo', 'stringfield': None}
```

Internally, we have the `Undefined` singleton as a placeholder to indicate the absence of a value.

```python
>>> m._data
{'intfield': Undefined, 'stringfield': None, 'defaultfield': 'foo'}
>>> 'intfield' in m
False
```

Attempting to access a nonexistent value through an attribute will raise an error that is a subclass of `AttributeError`.

There are two new parameters recognized by `import_loop`:

  * `init_values`: What to do with missing values

    * `True`: initialize to `None`
    * `False`: leave as `Undefined` (default)

  * `apply_defaults`: What to do with fields' defaults

    * `True`: honor them
    * `False`: ignore them (default)

Normally, `Model.__init__()` overrides both of these. Its arguments include `init=True`, which is just a shortcut for setting both `init_values` and `apply_defaults` at once when creating a new instance.

So, to avoid pre-populating the model, one can specify `init=False`:

```python
>>> m = M({'intfield': 1}, init=False)
>>> m.to_primitive()
{'intfield': 1}
>>> m._data
{'intfield': 1, 'stringfield': Undefined, 'defaultfield': Undefined}
```

Obviously, `init_values` and `apply_defaults` can also be set independently.

### Field defaults

Presently, Schematics turns `None` values into the fields' respective defaults on every import and even during validation, which IMO is plainly a bug.

As of this PR, these issues are fixed by virtue of

  * substituting defaults for *missing values only*
    (i.e., `None` in the input is now considered an explicit value that will *not* be replaced)
  * having `apply_defaults` turned off by default (as it's mainly useful during `__init__` only).

### Controlling what to export

There's a new option `export_level` that can appear in field definitions and model options. It supersedes `serialize_when_none`.

export_level | Undefined | None | Empty containers | Other values
-------------| --------- | ---- | ---------------- | ------------
ALL          | YES       | YES  | YES              | YES
DEFAULT      | NO        | YES  | YES              | YES
NOT_NONE     | NO        | NO   | YES              | YES
NONEMPTY     | NO        | NO   | NO               | YES
DROP         | NO        | NO   | NO               | NO

The default level is inherited from the base class options.

`serialize_when_none` remains as a (deprecated) synonym for `export_level`: `True => DEFAULT`, `False => NONEMPTY`.

`export_level` can also be passed as an argument to export methods, in which case it will override any field- and model-level settings. In this capacity, it replaces `print_none`, which is removed.

Missing values appear as `None` when exported.

```python
>>> m.to_primitive()
{'intfield': 1}
>>> m.to_primitive(export_level=ALL)
{'intfield': 1, 'stringfield': None, 'defaultfield': None}
```

### The model interface

`keys()`, `values()`, `items()`, and `__len__()` exclude unset fields.

By contrast, `atoms()` exposes the `Undefined` objects directly.

~~The subscription syntax also returns any `Undefined`s.~~

### Compatibility

Existing tests continue to pass unchanged except for

  * one test looking at the `del` statement, which only recently got fixed anyway
  * two tests that were passing the now-defunct `print_none` parameter to `export_loop`.

Breaking changes are mainly about fixing weird behaviors, most notably the habit of default values to reappear at every turn.

### DIFF
--- a/schematics/common.py
+++ b/schematics/common.py
@@ -1,5 +1,14 @@
+from .util import Constant
 
-NATIVE, PRIMITIVE = 0, 1
+
+NATIVE    = Constant('NATIVE',     0)
+PRIMITIVE = Constant('PRIMITIVE',  1)
+
+DROP      = Constant('DROP',       0)
+NONEMPTY  = Constant('NONEMPTY',   1)
+NOT_NONE  = Constant('NOT_NONE',   2)
+DEFAULT   = Constant('DEFAULT',   10)
+ALL       = Constant('ALL',       99)
 
 EMPTY_LIST = "[]"
 EMPTY_DICT = "{}"
--- a/schematics/exceptions.py
+++ b/schematics/exceptions.py
@@ -59,3 +59,14 @@ class StopValidation(ValidationError):
 class MockCreationError(ValueError):
     """Exception raised when a mock value cannot be generated."""
     pass
+
+
+class MissingValueError(AttributeError, KeyError):
+    """Exception raised when accessing an undefined value on an instance."""
+    pass
+
+
+class UnknownFieldError(KeyError):
+    """Exception raised when attempting to set a nonexistent field using the subscription syntax."""
+    pass
+
--- a/schematics/models.py
+++ b/schematics/models.py
@@ -13,6 +13,7 @@ from .datastructures import OrderedDict as OrderedDictWithSort
 from .exceptions import BaseError, ModelValidationError, MockCreationError
 from .types import BaseType
 from .types.serializable import Serializable
+from .undefined import Undefined
 
 try:
     unicode #PY2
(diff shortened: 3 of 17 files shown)


bintoro:
I've begun to question the wisdom of storing `Undefined` objects in `_data`. It might be more straightforward simply to leave out missing values altogether.

The `Undefined` singleton is only needed when we need a placeholder to represent the absence of a value. Notably, the transformation loops and the related functions need to be able to pass around a single value at a time, so `Undefined` is crucial there.

But the main API is not going to expose the placeholders, so why store them? It's easier to turn a missing `_data` item into an `Undefined` on the fly than to prune out the placeholders when they're not wanted.

PR #377, opened 2016-01-20
fixed 'cannot deepcopy' Issue on MultilingualStringType

devnoop:
There is an bug while copying a submodel with a MultilingualStringType in it. 
I got the error 'cannot deepcopy' caused while copying the compiled regex.

### DIFF
--- a/schematics/__init__.py
+++ b/schematics/__init__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
 
-version_info = ('1', '1', '1')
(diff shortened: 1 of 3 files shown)


bintoro:
Did you get the error with the current `development` branch? The root cause of the deepcopy problem was that types didn't implement `__deepcopy__`, but this was fixed on Dec 4.

devnoop:
I'm sorry. My fault. I used an older release to check it...

PR #379, opened 2016-01-26
Add NonDictModel class

mlyundin:
Please review. I have found that this is useful using schematics for validation of json data. According to json specification we can have not only 'dict' on top level.

### DIFF
--- a/schematics/models.py
+++ b/schematics/models.py
@@ -206,6 +206,15 @@ class ModelMeta(type):
     def fields(cls):
         return cls._fields
 
+class NonDictModelMeta(ModelMeta):
+    def __new__(mcs, name, bases, attrs):
+
+        klass =  ModelMeta.__new__(mcs, name, bases, attrs)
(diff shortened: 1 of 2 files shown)


bintoro:
What specification? The JSON specification certainly allows objects at the top level.

I can't help thinking there must a better solution to the problem you're facing. While this isn't mergeable as such, we can look into the actual issue if you can provide more information.

mlyundin:
From http://www.json.org/
In JSON, they take on these forms:
An object is an unordered set of name/value pairs.
An array is an ordered collection of values.

In my work I have faced with REST responses  with json data as array on top level

bintoro:
If you only wish to validate, say, a list of integers, you can simply use a standalone field:

```python
>>> field = ListType(IntType)
>>> field.validate([1, 2, "3", 4])
[1, 2, 3, 4]
```

You don't need a containing model to do this.

PR #380, opened 2016-01-26
Introduce MixedType

bintoro:
I've thought about PR #340 by @jmsdnns and decided to try to write a similar thing from scratch instead of adapting `PolyModelType`, since that one mostly consists of model-related machinery.

As explained in #340, saying "store value *x* with type *y*" is just not possible in the current Schematics architecture. However, as long as the type can be inferred from the value itself, it works great.

For now this thing is called `MixedType`.

#### A field that takes a numeric ID or a UUID
```python
>>> id_field = MixedType((IntType, UUIDType))
>>> result = id_field.to_native('ee5a16cb-0ee1-46bc-ac40-fb3018b1d29b')
>>> result
UUID('ee5a16cb-0ee1-46bc-ac40-fb3018b1d29b')
>>> id_field.to_primitive(result)
'ee5a16cb-0ee1-46bc-ac40-fb3018b1d29b'
>>> id_field.to_native('99999')
99999
```

#### A list of ints or floats
```python
numbers = ListType(MixedType((IntType, FloatType)))
>>> result = numbers.to_native(["2", "0.5", "123", "2.999"])
>>> result
[2, 0.5, 123, 2.999]
>>> [type(item).__name__ for item in result]
['int', 'float', 'int', 'float']
```

#### A dict of ints or a list of ints
```python
ints = MixedType((DictType, ListType), field=IntType)
>>> ints.to_native([1, 2, 3, 4])
[1, 2, 3, 4]
>>> ints.to_native(dict(a=1, b=2, c=3, d=4))
{'a': 1, 'c': 3, 'b': 2, 'd': 4}
```

#### Complex situations

All the previous cases rely on `to_native` to detect the intended type. You can get quite far with that method by exploiting the fact that the types are evaluated in the original order. The most restrictive type must therefore appear first. In the second example, `FloatType` would have caught all items had the order been `(FloatType, IntType)`.

For more complex cases, make a subclass of `MixedType` and declare the resolution logic there:

```python
class ShoutType(StringType):
    def to_primitive(self, value, context):
        return StringType.to_native(self, value) + '!!!'

class QuestionType(StringType):
    def to_primitive(self, value, context):
        return StringType.to_native(self, value) + '???'

class FancyStringType(MixedType):

    # Declare the possible types, either as classes or parameterized instances.
    types = (ShoutType, QuestionType)

    # Override the method that detects the correct type. Return the appropriate class.
    def resolve(self, value, context):
        if value.endswith('?'):
            return QuestionType
        elif value.endswith('!'):
            return ShoutType
```

```python
>>> FancyStringType().to_primitive("Hello, world!")
'Hello, world!!!!'
>>> FancyStringType().to_primitive("Who's a good boy?")
"Who's a good boy????"
```

A resolver function can be provided in the field definition too:

```python
fancy_field = MixedType((ShoutType, QuestionType), resolver=resolve_func)
```

Alternatively, edit the inner *types* and orchestrate their `to_native` methods so that the first successful conversion is guaranteed to be the right one.

#### Parameterizing types

Arguments destined for the inner types may appear directly in the `MixedType` definition and are passed on, much like with compound types.

```python
>>> FancyStringType(max_length=20).validate("What is the meaning of life?")
  ...
schematics.exceptions.ValidationError: [u'String value is too long.']
```

Since there can be quite dissimilar types that accept various sets of parameters, `MixedType` inspects all `__init__()` argument lists in the types' MROs in order to figure out which parameters go where.

```python
>>> field = MixedType((IntType, QuestionType), required=True, max_value=100, max_length=50)
>>> field.required
True
>>> field._types[IntType].required
True
>>> field._types[QuestionType].required
True
>>> field._types[IntType].max_value
100
>>> field._types[QuestionType].max_length
50
```

#### Parameterizing types separately

Just make instances as usual.

```python
>>> field = MixedType((IntType(min_value=1), FloatType(min_value=0.5)))
>>> field.validate(0)
  ...
schematics.exceptions.ValidationError: [u'Int value should be greater than or equal to 1.']
>>> field.validate(0.75)
>>>
```

### DIFF
--- a/schematics/types/__init__.py
+++ b/schematics/types/__init__.py
@@ -1,4 +1,5 @@
 from .base import *
+from .union import *
 from .net import *
 
 __all__ = [name for name, obj in globals().items() if isinstance(obj, TypeMeta)]
--- a/schematics/types/union.py
+++ b/schematics/types/union.py
@@ -1,0 +1,91 @@
+import inspect
+
+from ..common import *
+from ..datastructures import OrderedDict
+from ..exceptions import ConversionError
+from ..transforms import get_import_context, get_export_context
+from .base import BaseType
+
+
+def _valid_init_args(type_):
+    args = set()
+    for cls in type_.__mro__:
+        args.update(inspect.getargspec(cls.__init__).args[1:])
+        if cls is BaseType:
+            break
+    return args
+
+def _filter_kwargs(valid_args, kwargs):
+    return dict((k, v) for k, v in kwargs.items() if k in valid_args)
+
+
+class UnionType(BaseType):
+
+    types = None
+
+    MESSAGES = {
+        'convert': u"Couldn't interpret '{0}' value as any of {1}.",
+    }
+
+    _baseclass_args = _valid_init_args(BaseType)
+
+    def __init__(self, types=None, resolver=None, **kwargs):
+
+        self._types = OrderedDict()
+        types = types or self.types
+        if resolver:
+            self.resolve = resolver
+
+        for type_ in types:
+            if isinstance(type_, type) and issubclass(type_, BaseType):
+                type_ = type_(**_filter_kwargs(_valid_init_args(type_), kwargs))
+            elif not isinstance(type_, BaseType):
+                raise TypeError("Got '%s' instance instead of a Schematics type" % type_.__class__.__name__)
+            self._types[type_.__class__] = type_
+            self.typenames = tuple((cls.__name__ for cls in self._types))
+
+        super(UnionType, self).__init__(**_filter_kwargs(self._baseclass_args, kwargs))
+
+    def resolve(self, value, context):
+        for field in self._types.values():
+            try:
+                value = field.convert(value, context)
+            except ConversionError:
+                pass
+            else:
+                return field, value
+        return None
+
+    def _resolve(self, value, context):
+        response = self.resolve(value, context)
+        if isinstance(response, type):
+            field = self._types[response]
+            try:
+                response = field, field.convert(value, context)
+            except ConversionError:
+                pass
+        if isinstance(response, tuple):
+            return response
+        raise ConversionError(self.messages['convert'].format(value, self.typenames))
+
+    def convert(self, value, context=None):
+        context = context or get_import_context()
+        field, native_value = self._resolve(value, context)
+        return native_value
+
+    def validate(self, value, context=None):
+        field, _ = self._resolve(value, context)
+        return field.validate(value, context)
+
+    def _export(self, value, format, context=None):
+        field, _ = self._resolve(value, context)
+        return field._export(value, format, context)
+
+    def to_native(self, value, context=None):
+        field, _ = self._resolve(value, context)
+        return field.to_native(value, context)
+
+    def to_primitive(self, value, context=None):
+        field, _ = self._resolve(value, context)
+        return field.to_primitive(value, context)
+
--- a/schematics/validate.py
+++ b/schematics/validate.py
@@ -101,7 +101,8 @@ def get_validation_context(**options):
         'partial': False,
         'strict': False,
         'convert': True,
-        'validate': True
+        'validate': True,
+        'new': False,
     }
     validation_options.update(options)
     return Context(**validation_options)
--- a/tests/test_union_type.py
+++ b/tests/test_union_type.py
@@ -1,0 +1,113 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import print_function, division
+
+import pytest
+import uuid
+
+from schematics.common import *
+from schematics.exceptions import *
+from schematics.models import Model
+from schematics.types import *
+from schematics.types.compound import *
+from schematics.types.serializable import Serializable
+
+
+def test_id_or_uuid():
+
+    id_field = UnionType((IntType, UUIDType))
+
+    result = id_field.convert('ee5a16cb-0ee1-46bc-ac40-fb3018b1d29b')
+    assert result == uuid.UUID('ee5a16cb-0ee1-46bc-ac40-fb3018b1d29b')
+    assert type(id_field.to_primitive(result)) is str
+
+    result = id_field.convert('99999')
+    assert result == 99999
+    assert type(result) is int
+    assert type(id_field.to_primitive(result)) is int
+
+
+def test_list_of_numbers():
+
+    numbers = ListType(UnionType((IntType, FloatType)))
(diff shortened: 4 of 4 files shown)


quasiben:
What's the status of merging this or PR #340 ?

bintoro:
The code here is fairly robust as is, so it's pretty much mergeable once I rebase.

I'd like to come up with a better name for the type, though. `MixedType` is far from ideal.

With the advent of the `typing` module in Python 3.5, I've grown sympathetic to `UnionType` suggested in #340. The semantics of `typing.Union` is that of a union type, i.e., "one type from a set of *n*>=1".

Now that the concept of a union type has landed in the standard library, the default choice is to use the same nomenclature. If anyone disagrees, please voice your alternative suggestion **now**. This PR has been sitting around long enough.

As for future plans, I intend to deprecate `PolyModelType` and make its subclass resolution logic part of `ModelType` itself. Now that we have a generic union type, there's no reason for `PolyModelType` to exist.

quasiben:
Is it reasonable to assume that merging this PR will result in #340 being closed ?

bintoro:
@quasiben Yes. #340 remains at a draft stage, and there is probably little reason to pursue that avenue any longer, given that this PR has accomplished what #340 set out to do. Is there something in the design of #340 that has not been adequately addressed here?

quasiben:
I may be missing something in your examples above but I would've thought `MixedType` could work with `StringType`/`IntType`.  For example, the following should return a `ConversionError`, no ?

```
from schematics.types import MixedType, IntType, StringType
from schematics.types.compound import ListType

things = ListType(MixedType((IntType, StringType(choices=['a', 'b']))))
things.to_native([2, 'a', 'foo'])
```

quasiben:
And in response to your previous question, no, I was just curious what your plans were.  Also, thanks!  Schematics is pretty great!!!

bintoro:
Conversion is about type checking and coercion, but `choices` is a validation concern:

```python
>>> things.validate([2, 'a', 'foo'])
...
schematics.exceptions.CompoundError: {2: ValidationError("Value must be one of ['a', 'b'].")}
```

The difference between conversion and validation is one of the things that should be spelled out more clearly in the documentation. `ConversionError` is a subclass of `TypeError`, while `ValidationError` is a subclass of `ValueError`. Substitute `['foo']` for `'foo'` and you'll see it fail at the conversion stage.

quasiben:
Thanks for the clarification.  

In the above `validate` case `2` is throwing the `ValidationError` but it shouldn't be, right?  2 is a valid `IntType` and the MRO should be `IntType` then `StringType`

quasiben:
If there a more preferable communication channel I'm happy to switch.  Populating this PR may not be desirable

bintoro:
> In the above `validate` case 2 is throwing the `ValidationError` but it shouldn't be, right? 2 is a valid `IntType` and the MRO should be `IntType` then `StringType`

Just a coincidence :smile: The number in the error message is actually the index of the offending item, which has been displayed since #374.

Discussing here is fine; we don't get too much traffic.

quasiben:
Ah!!!  Can I suggest we change to the value or a tuple of (index, value) ?

bintoro:
Errors are aggregated in a dictionary structure, where they are keyed by field name (for models) or by index (for `ListType`). Including the value in the key is not directly possible because keys must be hashable.

That said, the main thing holding the 2.0 release back is that I want to improve the error system further. Specifically, we should have a universal mechanism for accessing details about each atomic error throughout the error's lifecycle. Right now the details are embedded in a string — not good.

In #374, purely textual error messages were replaced with `ErrorMessage` objects that can host structured information about each error. What remains to be done is to make the bundled types actually utilize this feature.

An important application of all this is localization of error messages. I intend to open a dedicated issue for it at some point.

quasiben:
Happy to help if you want to assign some introductory work

quasiben:
As a final question on the `MixedType` issue I wanted to step things up a bit more.  The below is currently failing:

```python
from schematics.types import MixedType, IntType, StringType
from schematics.types.compound import ListType, ModelType, PolyModelType

class Foo(Model):
    a = StringType()
    b = IntType()

Foo(dict(a='twelve', b=1))

things = ListType(MixedType((IntType, from schematics.models import Model
from schematics.types import MixedType, IntType, StringType
from schematics.types.compound import ListType, ModelType, PolyModelType

class Foo(Model):
    a = StringType()
    b = IntType()

Foo(dict(a='twelve', b=1))

things = ListType(MixedType((IntType, PolyModelType(Foo), StringType(choices=['a', 'b']))))
things.validate([2, 'a', 'b'])

things.validate([2, 'a', 'b', dict(a='twelve', b=1)])
things.validate([2, 'a', 'b', dict(a='twelve', b=1, c=2)])

class Profile(Model):
    thing = things

pp = Profile(dict(thing=[2, 'a', 'b', dict(a='twelve', b=1)]))(Foo), StringType(choices=['a', 'b']))))
things.validate([2, 'a', 'b'])

things.validate([2, 'a', 'b', dict(a='twelve', b=1)])
things.validate([2, 'a', 'b', dict(a='twelve', b=1, c=2)])

class Profile(Model):
    thing = things

pp = Profile(dict(thing=[2, 'a', 'b', dict(a='twelve', b=1)]))

Exception: Cannot export: <type 'dict'> is not an allowed type
```

Originally, `PolyModelType` was `ModelType` but this lead to an early error

bintoro:
@quasiben There are a couple of things going on here.

I'll start by mentioning that operating on standalone fields (that are not part of any model) is not really well-established usage. Up until now, standalone fields have been used in test cases but little more.

Recently, however, I've been adding bits and pieces to make it possible to use a standalone type as the root element. I can definitely see value in that usage. Sometimes you just need to coerce or validate a list of items, for instance. The items themselves may be complex, but there just isn't a model-like structure at the root.

For the moment, this only has experimental support, since there are no comprehensive tests yet.

Anyway, congrats — you found 2 bugs. One had to do with what I explained above; a standalone `ListType` would not import a nested `ModelType`. The other was a bona fide regression in `MixedType`.

Both have now been fixed in the `mixed-type` branch, and your sample code should work. But note that `MixedType` has now been renamed `UnionType`.

quasiben:
Any idea when this will be merged ?

bintoro:
How about now? :)

quasiben:
Awesome!  Thanks so much!!!

quasiben:
I suppose the next question is if there is a scheduled release in the near future ? (Also, happy to help here)

PR #387, opened 2016-02-11
Add example to DictType

charlax:
It is unclear if `type` is (1) a Python `type` (2) a schematics type defined in `schematics.types`. Adding an example clarifies which is which to avoid exceptions such as `AttributeError: 'str' object has no attribute 'owner_model'`.

### DIFF
--- a/schematics/types/compound.py
+++ b/schematics/types/compound.py
@@ -149,6 +149,12 @@ class ModelType(CompoundType):
 class ListType(CompoundType):
     """A field for storing a list of items, all of which must conform to the type
     specified by the ``field`` parameter.
+
+    Use it like this:
+
+    .. code-block:: python
+        ...
+        categories = ListType(StringType)
     """
 
     def __init__(self, field, min_size=None, max_size=None, **kwargs):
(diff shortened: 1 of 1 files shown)


kracekumar:
@charlax Thanks. `ListType` also needs an example, can you add ?

schematics/types/compound.py
```diff
@@ -244,6 +244,13 @@ def _export(self, list_instance, format, context):
 class DictType(CompoundType):
     """A field for storing a mapping of items, the values of which must conform to the type
     specified by the ``field`` parameter.
+    
+    Use it like this:
+    
+    .. code-block:: python
```
bintoro: The `code-block` directive is unnecessary. A simple `::` marker is preferred (`Use it like this::`).

charlax: Yes - but only if Python is the default highlighted language. I usually prefer to be explicit. Let me know if you feel strongly about this and I'll make the change.

charlax:
Added the example.

bintoro:
Thanks! Merged via 0ead9c19a9e48de13d7d3967c3e9781fac2f45c0.

PR #389, opened 2016-02-12
Fix mocked DateTimeType timezone behavior

### LINKED ISSUE #388: When used with get_mock_object, DateTimeType returns naive datetime
Fairly straightforward to reproduce:

```
In [1]: from schematics.types import DateTimeType

In [2]: from schematics.models import Model

In [11]: class Toaster(Model):
    created_at = DateTimeType(required=True)

In [15]: Toaster.get_mock_object().created_at.tzinfo

In [16]:
```

Since people might have assumed this would always be the case, perhaps an extra constructor arguments `default_timezone` would be useful? This would make sure that only timezone-aware datetime are returned. It would use the timezone on the parsed datetime if it's there, or the default one.

Thoughts?

charlax:
Fixes #388.


charlax:
There's an issue with the rebase - will have a look.

charlax:
Fixed the rebase.

schematics/types/base.py
```diff
@@ -1,39 +1,36 @@
 # -*- coding: utf-8 -*-
 
 import copy
-import math
```
charlax: I'm also cleaning up unused imports

schematics/types/base.py
```diff
@@ -1,39 +1,36 @@
...
 from ..common import *
-from ..datastructures import Context
 from ..exceptions import ConversionError, ValidationError, StopValidationError
 from ..undefined import Undefined
 from ..util import listify
 from ..validate import prepare_validator, get_validation_context
 
 try:
-    from string import ascii_letters # PY3
+    from string import ascii_letters  # PY3
```
charlax: And applying pep8.

schematics/types/base.py
```diff
@@ -791,6 +787,25 @@ def _mock(self, context=None):
...
+            mocked = mocked.replace(tzinfo=self.offset_timezone(hours=hours))
+
+        elif self.tzd == 'utc':
+            # Always return utc.
+            mocked = mocked.replace(tzinfo=self.UTC)
+
+        # If tzd == 'reject', we return a timezone-naive datetime.
+
+        return mocked
+
```
bintoro: Looks correct but doesn't take `convert_tz` or `drop_tzinfo` into account. Would you mind adding those?
* When `convert_tz` is true, the datetime can only be naive or UTC.
* When `drop_tzinfo` is true, the datetime can only be naive.

charlax:
Updated the PR with the comment. The code could have been more DRY but I decided to focus on readability.

charlax:
Also, the tests will be fixed after we agree on the behavior first.

charlax:
@bintoro can you review? Tests & behavior have been fixed.

bintoro:
I really appreciate the effort, but unfortunately the code is still not accounting for all possible option combinations. The method is also growing rather unwieldy in relation to what it's meant to do.

For these reasons I decided to opt for a more straightforward and functionally complete approach committed [here](https://github.com/schematics/schematics/commit/2110749141613a411c47263f2105c5c5adb4cd0d#diff-d843f6ce825bc3ed6e2e66d597044f36R762).

Just a heads-up: while reviewing the code, I realized there was a bug in the `validate_tz()` method. Setting `tzd='utc'` should *not* automatically imply `convert_tz=True`, but the validator behaved as if it did. This was also fixed in the commit I linked to.

Issue #375, opened 2015-12-28
RFC: Model API


bintoro:
Some thoughts on the Model interface below. I'd like to hear what others think.

##### Dealing with inheritance

As I was rewriting `Model.__eq__()` to handle self-referential structures, I noticed it performs an `isinstance` check on the other object instead of a strict type equivalence check. The problem is, it goes on to ignore extra fields on subclasses:

```python
>>> class Base(Model):
...     field1 = IntType()
... 
>>> class Derived(Base):
...     field2 = IntType()
... 
>>> Base({'field1': 1}) == Derived({'field1': 1, 'field2': 123})
True
```

I don't think this is fine at all. The big question is when (if ever) should the two be considered equivalent, given that `field2` will never exist on a `Base` instance. I have tentatively solved this by considering both `None` and `Undefined` on the subclass to satisfy equivalence with a nonexistent field on the base.

A slightly more rigorous approach would be to maintain a dirty bit on all fields and only equate an empty value with a missing field *if the empty value was there to begin with*, but this may be more trouble than it's worth.

A comparison algorithm that works with inheritance is definitely a useful thing to have. I'm just wondering if the `==` operator is the right place to put it. Furthermore, there should be a strict comparison available, too, as a shortcut for `obj1.__class__ is obj2.__class__ and obj1 == obj2`.

##### No type checks on values

Since `Model.__eq__()` only does a basic Python equivalence check to compare the individual fields, it's perfectly possible to have two instances that compare equal, one of which passes validation while the other doesn't:

```python
>>> class M(Model):
...     boo = BooleanType()
... 
>>> m1 = M()
>>> m1.boo = 1
>>> 
>>> m2 = M()
>>> m2.boo = 1.0
>>> 
>>> m1 == m2
True
>>> 
>>> m1.validate() # passes
>>> m2.validate() # fails
```

Unfortunately, correcting this would involve making copies of both instances, running them through `to_native()` and comparing the coerced values, adding quite a bit of overhead.

I'm thinking this could be addressed through documentation. The pitfall only exists if a value is set by attribute access using a wrong/unknown type, which in unsafe anyway; it should always be followed by `to_native()` or `validate()`.

##### What to iterate over

The present behavior is to iterate over field names. I believe it would be more useful to iterate over name–value pairs.

##### Handling undefineds

The way the Undefined PR affects this is that any fields with an undefined value are skipped. The same goes for iterating over `keys()`, `values()`, and `items()`.

[...]

PR #390, opened 2016-02-12
Add support for arbitrary field metadata.

chadrik:
Hi,
Here's a pull request for the metadata/annotations feature discussed in #333

It's dead simple.  I went with the name `metadata` instead of `annotations` because I think `metadata` is more self-explanatory, and I feel that `annotations` implies strings only.

### DIFF
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -166,7 +166,7 @@ class BaseType(TypeMeta('BaseTypeBase', (object, ), {})):
     def __init__(self, required=False, default=Undefined, serialized_name=None,
                  choices=None, validators=None, deserialize_from=None,
                  export_level=None, serialize_when_none=None,
-                 messages=None, **kwargs):
+                 messages=None, metadata=None, **kwargs):
         super(BaseType, self).__init__()
 
         self.required = required
@@ -184,6 +184,7 @@ class BaseType(TypeMeta('BaseTypeBase', (object, ), {})):
         self._set_export_level(export_level, serialize_when_none)
 
         self.messages = dict(self.MESSAGES, **(messages or {}))
+        self.metadata = metadata or {}
         self._position_hint = next(_next_position_hint)  # For ordering of fields
 
         self.name = None
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -150,7 +150,13 @@ class BaseType(TypeMeta('BaseTypeBase', (object, ), {})):
         class. A metaclass will merge all the `MESSAGES` and override the
         resulting dict with instance level `messages` and assign to
         `self.messages`.
-
+    :param metadata:
+        Dictionary for storing custom metadata associated with the field.  
+        To encourage compatibility with external tools, we suggest these keys 
+        for common metadata:
+        - *label* : Brief human-readable label
+        - *description* : Explanation of the purpose of the field. Used for 
+          help, tooltips, documentation, etc.
     """
 
     MESSAGES = {
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -436,3 +436,8 @@ def test_geopoint_range():
 
     with pytest.raises(ValidationError) as ve:
         geo.validate((28.2342323, 181.123141))
+
+def test_metadata():
+    metadata = {'description': 'amazing', 'extra': {'abc': 1}}
+    assert StringType(metadata=metadata).metadata == metadata
+    assert StringType().metadata == {}


bintoro:
Looks pretty good to me! I have hesitations about the name `metadata`, though. Programs might have uses for the container that go beyond descriptive properties. A database column, for instance, is not really metadata IMO.

For reference, the import and export loops have a proprietary data bag called `app_data`, but I can't come up with anything useful following that precedent.

Applications using this feature are likely going to pass the argument to many, perhaps most, of their fields, so the shorter the better. Something that occurred to me is simply `ext` for extended or extension attributes. Pro: short; usage-agnostic; could be used elsewhere as well. Contra: usage not self-evident.

Any ideas? Basically the name should convey "custom attributes go here".

chadrik:
I think `ext` is a bit too cryptic.  Here are some other ideas that meet the brevity requirement:

- `attrs`:  it's actually a dictionary, so maybe misleading?
- `data`: pretty generic, but then again, so is its purpose
- `extra`: my favorite

and one more inspired by `ext` but a bit more explicit:
- `external`

bintoro:
Mm-hmm. I tried to push for `attrs` in #333. Now I'm thinking it's too vague because nothing indicates it's for proprietary attributes. The same goes for `data`.

I'm surprised to hear `ext` seems cryptic; I thought it was pretty standard for "extension", "extended", etc. But `extra` is definitely a good alternative. Anyone else want to chime in?

In #333, I envisioned that the extras would be exposed as attributes on the container object: `somefield.extra.description`. (This is similar to the `context` object present during import/export.) What are your thoughts on this?

chadrik:
> I'm surprised to hear `ext` seems cryptic; I thought it was pretty standard for "extension", "extended", etc. 

It's growing on me a little bit, but I still prefer more explicit names.

> In #333, I envisioned that the extras would be exposed as attributes on the container object: `somefield.extra.description`. (This is similar to the context object present during import/export.) What are your thoughts on this?

I'm concerned that this introduces all of the issues regarding name clashes and dictionary methods that we've talked about in #375.  It seems acceptable and even intuitive to me that a container that holds *custom, external* data is left completely as-is, but I also see the appeal of attr-based access.

Consider this.  Stylistically a lot of people will probably prefer to declare their extra data like this:

```python
class Person(Model):
    first_name = StringType(
        required=True,
        extra=dict(
            title='First Name',
            description="The person's first name")
        )
    age = IntType(
        extra=dict(
            description="The person's age",
            as_sql_col='NUMBER')
        )
```

So, I think a simple solution is for schematics to provide an attr-dict util class that users can use if they like:

```python
class Person(Model):
    first_name = StringType(
        required=True,
        extra=AttrDict(
            title='First Name',
            description="The person's first name")
        )
    age = IntType(
        extra=AttrDict(
            description="The person's age",
            as_sql_col='NUMBER')
        )
```

I've also been thinking about your concerns about the term `metadata`.  I could be wrong, but I don't think that most developers feel strongly that metadata must be purely descriptive.  What we're talking about is data related to other data (the field), and I don't think that someone who wants to associate a sql column with a `schematics` field will be confused or discouraged when they spot "metadata" in the code or docs.  Also, not that it's a strong argument, but "metadata" is the term that [`marshmallow`](https://marshmallow.readthedocs.org/en/latest/) uses for the same purpose.

My 2 c.  I'll take a poll of the devs here at work to see what they think.

chadrik:
In case it helps the four devs that I forwarded this discussion to voted for `metadata`.

nrusch:
I'm one of the developers @chadrik mentioned, and I just wanted to add my voice in support of this, and more specifically, in favor of the `metadata` keyword.

Within development circles, the concept of "metadata" is pretty universally understood at this point, and I think the value of a keyword that is effectively self-documenting outweighs any of the potential advantages of choosing a shorter alternative.

> Programs might have uses for the container that go beyond descriptive properties. A database column, for instance, is not really metadata IMO.

The only thing "metadata" really says is that its contents have no direct impact on the functionality of the entity it is attached to. Whether those contents affect the functionality of a containing **program** is irrelevant. A database column, for instance, is absolutely metadata.

Thanks.

bintoro:
Okay, you've convinced me.

One last question before I merge. Now that we've agreed on *metadata* in principle, do you guys also prefer `metadata` over just `meta`? I don't think the shortened variant ever came up, even though it's widely used as well. I'm fine with either.

nrusch:
Personally, while I don't mind `meta`, I would say I prefer the specificity of `metadata`. Even at twice the character count, it's far from the longest keyword name for fields.

PR #399, opened 2016-03-05
Test for Model.__init__ context passing to types

sheilatron:
This test is to verify that a context dict, passed into model initialization, is available for use by custom data types, with the context dict keys converting to context object attributes.

### DIFF
--- a/tests/test_model_type.py
+++ b/tests/test_model_type.py
@@ -186,6 +186,7 @@ def test_specify_model_by_name():
 
 def test_model_context_pass_to_type():
     from schematics.types import BaseType
+    from schematics.datastructures import Context
 
     class CustomType(BaseType):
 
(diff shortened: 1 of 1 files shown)


tests/test_model_type.py
```diff
@@ -182,3 +182,25 @@ class M(Model):
...
+            return str(value) + suffix
+
+        def to_primitive(self, value, context=None):
+            suffix = context.suffix
+            return value[:-len(suffix)]
+
+    class Thing(Model):
+        x = CustomType()
+
+    context = {'suffix': 'z'}
```
sheilatron: Hi @bintoro note that I'm passing a dict here for the context. An actual Context instance didn't work out well, which surprised me.

bintoro:
All right, let's tackle this.

First I should mention that custom attributes are kind of meant to go in the `app_data` parameter, which will be made available during the loop as `context.app_data`. Using `context` is mostly fine, though — it's just that there is currently no namespacing for Schematics vs. private options.

The reason the code fails with an actual `Context` object is that the exact same object is being used first for model instantiation (import) and later export. This is a huge no-no, and it's really good this came up because we need to make sure this is well documented. `Context` instances encapsulate all kinds of options regarding the import/export process, and reusing them (even when it *does* work) will lead to unpredictable behavior unless you're aiming to replicate the settings on purpose.

However, passing a **freshly constructed** `Context` instance instead of a `dict` is perfectly fine and does work. It's even necessary in cases where the initiating code needs to retain a reference to the context after it's been populated in `import_loop` or `export_loop`.

All of that said, the PR itself is good. While we're at it, would you mind replicating it to test the `app_data` parameter as well?

sheilatron:
Thanks @bintoro. I should have some time to add those tests tomorrow, and look into how app_data works.

sheilatron:
Um...I was busier than expected over the past few days, but should be able to get to this soon.

sheilatron:
Hi @bintoro , I've added the tests you suggested, along with a TODO comment suggesting a more friendly error when the same Context instance is reused improperly.

sheilatron:
I'm not sure why the pypy3 test is failing; I didn't touch the module that it's failing on (test_models.py). The only one I modified was test_model_type.py.

bintoro:
Seems to have been a temporary glitch. I ran the Travis build again and it passed without issue. Thanks!

PR #400, opened 2016-03-05
Expose compound types to `schematics.types` interface

jamesob:
Small change; just noticed that you can't reference `types.DictType` when doing `from schemata import types`, which is a bummer.

### DIFF
--- a/schematics/models.py
+++ b/schematics/models.py
@@ -15,7 +15,7 @@ from .transforms import (
     flatten, expand
 )
 from .validate import validate, prepare_validator
-from .types import BaseType
+from .types.base import BaseType
 from .types.serializable import Serializable
 from .undefined import Undefined
 
--- a/schematics/types/__init__.py
+++ b/schematics/types/__init__.py
@@ -1,4 +1,5 @@
 from .base import *
 from .net import *
+from .compound import *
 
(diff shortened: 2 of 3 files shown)


bintoro:
The reason `types.compound` isn't imported right away is a circular dependency from `types.compound` to `Model` and back to `types` again.

One strategy to cope with this would be to patch the `types` subpackage from `schematics/__init__.py` once everything else has finished importing. Thoughts?

In principle I'm :+1: on this.

jamesob:
@bintoro apologies for the delay in reply. I've pushed one approach which is ugly but explicit. Nesting imports seems slightly less tricky than an ordered patching, but I'm amenable to either.

bintoro:
I'm sorry but imports inside methods are a no-go, particularly `pre_setattr`. Even though the modules themselves are cached, invoking the import statement is still relatively slow.

The monkeypatch approach I envisioned is quite straightforward: b35f3236432f23bc3c5625d1adec694d74001e85

Unless someone can see a problem with that, I'll merge it into mainline.

bintoro:
Compound types are now available directly under `schematics.types` (0bd4713).

jamesob:
:+1: thanks for the fix.

PR #407, opened 2016-04-27
Change conversion rules for booleans

### LINKED ISSUE #315: StringType() allows casting from int to str
Hi, 

I noticed that StringType defines allow_casts = (str, int).
I wonder what the motivation is for this. I was quite surprised that
a StringType() field silently casts int input to str.

I can always make a subclass that overrides the allow_casts class variable though.

Thanks for your help!

bintoro:
- `StringType` will reject `bool`.
- `IntType` will upcast `bool` to `int`.

***

The fact that the Python `bool` is a subclass of `int` is more a historical accident than a well-founded principle of the type system. Fundamentally, booleans have nothing to do with integers. Therefore I believe special-casing `bool` is entirely warranted.

Resolves #315.

### DIFF
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -382,6 +382,8 @@ class StringType(BaseType):
                     return str(value, 'utf-8')
                 except UnicodeError:
                     raise ConversionError(self.messages['decode'].format(value))
+            elif isinstance(value, bool):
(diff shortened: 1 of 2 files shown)


lkraider:
Looks good to me.

PR #409, opened 2016-04-29
RFC: Validate should either pass self or discover class methods

alexhayes:
As [outlined in the docs](https://github.com/schematics/schematics/blame/development/docs/usage/validation.rst#L125), a custom `validate_*` method receives `self`.

However this isn't actually the case, it receives the class (as tested in `test_validate_methods_passed_self`).

This would only be a documentation issue if methods decorated with `@classmethod` where actually discovered, but they aren't, as proven in my test `test_validate_discovers_classmethod`.

In order to fix this, which direction should be taken?

After looking the code I would think the intention is that `validate_*` methods should be decorated with `@classmethod` and be discoverable (as `self` is relatively useless, and potentially "dangerous", if being called within an `__init__`)

### DIFF
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -243,6 +243,18 @@ def test_validate_methods_passed_self():
     Foo({'foo': u'Bar'}).validate()
 
 
+def test_validate_discovers_classmethod():
+    class Foo(Model):
+        foo = StringType()
+
+        @classmethod
+        def validate_foo(cls, data, value, context):
+            raise ValidationError(u"I'm a classmethod that should be discovered.")
+
+    with pytest.raises(DataError):
+        Foo({'foo': u'Bar'}).validate()
+
+
 def test_basic_error():
     class School(Model):
         name = StringType(required=True)
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -233,6 +233,15 @@ def test_multi_key_validation_fields_order():
         Signup({'name': u'Brad'}).validate()
 
 
+def test_validate_methods_passed_self():
+    class Foo(Model):
+        foo = StringType()
+
+        def validate_foo(self, data, value, context):
+            assert self.__class__.__name__ == 'Foo'
+
+    Foo({'foo': u'Bar'}).validate()
+
 
 def test_basic_error():
     class School(Model):


bintoro:
Yeah, the docs have an incorrect depiction of the situation; the first argument is always the class.

This is because you can run validation on a plain dict without ever creating an instance. So, there might not be an instance to pass. (It *would* be possible to pass an instance when one exists, though, but this is not currently implemented.)

Essentially, the validators are implicit classmethods. It turns out that the discovery mechanism actually gets confused by a `classmethod` decorator. I've put together [a patch](https://github.com/schematics/schematics/compare/classmethod-validators). Even though the decorators are redundant, this allows them be used.

For someone not familiar with Schematics, it may be confusing to see `cls` as the first parameter when there's no `classmethod` around, so I suppose it makes sense that explicit decorators can be added for clarity.

Before I merge the patch, I'd like to ask for some opinions here:

* Does everybody agree with allowing the decorator, or is there a downside to having two different conventions?
* Would there be any point in passing an instance to the validator when one exists? If yes, should the decision be based on
    * whether the validator has been designated as a classmethod or
    * whether the first argument is `cls` or `self`?

What about when the validator wants `self` but we don't have an instance? It looks like this could get confusing very fast.

alexhayes:
> - Does everybody agree with allowing the decorator, or is there a downside to having two different conventions?

I think if you are going to pass the class you should definitely support the decorator. Calling a variable `self` when the convention dictates that it's not actually `self` is misleading. Also, given if the validate is run at time of instantiation when values haven't actually been set on `self` makes little sense. The other reason to declare the decorator is for linting purposes.

If you were going to pass `self`, then as a user I would expect it to be filled with values which can be validated. This in my opinion would be much nicer from an API point of view, but arguably not as flexible for testing purpose. For instance;

```python
class Foo(Model):
    foo = StringType()
    bar = BooleanType()

    def validate_foo(self, value):
        if self.bar and len(value) == 0:
            raise ...
```

Testing this is obviously a little more involved that testing the class based method that just takes a `dict` and value.

However, in saying this I think there should be one convention. If the intention is that they are class methods, then they should be class methods with a decorator, they shouldn't sometimes receive the class and sometimes the instance.

bintoro:
> Calling a variable `self` when the convention dictates that it's not actually `self` is misleading.

It is, but this is just an issue with the docs, which are old and buggy anyway. They need a thorough update before we can tag 2.0.

The name of the initial parameter is of no consequence as far as Schematics is concerned. You are free to name it `cls`, and indeed that is the correct usage IMO, because they behave as de facto classmethods. I have no idea why the docs have `self` there.

> If the intention is that they are class methods, then they should be class methods with a decorator, they shouldn't sometimes receive the class and sometimes the instance.

This sentence conflates the decorator issue and the instance vs. class issue, so I'm not sure what it is you're proposing.

I can merge the patch that adds support for the `classmethod` decorator, but I have no desire to introduce a requirement to use it. Implicit special signatures are not unheard of, `__new__` for instance.

alexhayes:
>> The name of the initial parameter is of no consequence as far as Schematics is concerned. You are free to name it cls, and indeed that is the correct usage IMO, because they behave as de facto classmethods. I have no idea why the docs have self there.

Yes of course, apart from conventions and linters complaining.

>> If the intention is that they are class methods, then they should be class methods with a decorator, they shouldn't sometimes receive the class and sometimes the instance.
> This sentence conflates the decorator issue and the instance vs. class issue, so I'm not sure what it is you're proposing.

Sorry, I wasn't inferring that you force the use of class method, just that it is clear that it definitely *should* be a class method.

My vote is that the patch is merged and docs are updated (which I'd be happy to contribute to in a different PR unless a major rewrite is in progress).

bintoro:
The patch and your testcase have been committed (920dfec) and included in 2.0.0.a1.

alexhayes:
Great thanks!

PR #410, opened 2016-04-29
RFC: Compound type values shared across instances when default supplied

alexhayes:
It appears that if a default is supplied to either `DictType` or `ListType` then the value on an actual model instance is shared across instances.

Is this the intention?

### DIFF
--- a/tests/test_dict_type.py
+++ b/tests/test_dict_type.py
@@ -159,3 +159,16 @@ def test_key_type():
     assert d == {
         "categories": {1: {"slug": "math"}}
     }
+
+
+def test_dict_default_not_shared():
+    class Foo(Model):
+        foo = DictType(IntType, default=dict())
+
+    one = Foo()
+    one.foo['a'] = 3
+
+    two = Foo()
+
+    assert one.foo['a'] == 3
+    assert 'a' not in two.foo
--- a/tests/test_list_type.py
+++ b/tests/test_list_type.py
@@ -236,3 +236,15 @@ def test_mock_object_with_model_type():
         age = IntType(required=True)
 
     assert isinstance(ListType(ModelType(User), required=True).mock()[-1], User)
+
+
+def test_list_default_not_shared():
+    class Foo(Model):
+        foo = ListType(IntType, default=[])
+
+    one = Foo()
+    one.foo.append(1)
+
+    two = Foo()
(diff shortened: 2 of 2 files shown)


bintoro:
It's the same phenomenon as when you assign a mutable container as a default value for a function parameter — it'll get evaluated only once, and the result is shared across invocations.

The right way to do this is to assign a callable to `default`, like `default=dict` without parentheses. It will then be invoked anew for every instance.

alexhayes:
> It's the same phenomenon as when you assign a mutable container as a default value for a function parameter — it'll get evaluated only once, and the result is shared across invocations.

Of course, but from the users point of view it's completely not what is expected.

For instance, if I want to have a default `dict` with values I must do this;

```python
foo = DictType(IntType, default=lambda: dict(eggs=1, sausage=5))
```

That seems a little odd to me and somewhat un-pythonic.

I would also question a situation where you'd actually want to share that default `dict` across instances of models, seems a very unlikely use case to me, if you want to share a variable across all instances there are much better alternatives that aren't going to cause the confusion that this approach will.

Would you consider a deep copy of the default on instantiation?

bintoro:
I'm not sure I agree on the "completely unexpected" part. Have you never used [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict) of the standard library or SQLAlchemy column defaults, both of which take factory functions?

There's a reason factory functions are used in situations like these. First, it's impossible to know what kind of copy is appropriate. Second, sometimes you want a default value that is not static, such as a timestamp. If you have an elegant solution to these issues, I'm happy to consider it.

alexhayes:
Agreed, not debating the support of factory functions, however I guess I was just expecting a copy would be created simply because I can't think of a use case where I would want one model to change the data on another. After further research it seems that other libraries have left this decision up to the user as schematics has, so I stand corrected on the "completely unexpected" - apologies.

Perhaps a simple addition to the docs is in order for the compound field types simply reminding users of this behaviour. Django's [JSONField](https://docs.djangoproject.com/en/1.9/ref/contrib/postgres/fields/#jsonfield) states the following;

> If you give the field a `default`, ensure it’s a callable such as `dict` (for an empty default) or a callable that returns a dict (such as a function). Incorrectly using `default={}` creates a mutable default that is shared between all instances of `JSONField`.

PR #425, opened 2016-05-25
When testing if a values is None or Undefined, use 'is'.

chadrik:
`value in (None, Undefined)` invokes `value.__eq__`, which may have side effects.


bintoro:
You beat me to it — I had made the same change locally. It's a noticeable speedup even when `value.__eq__` doesn't do anything fancy.

Issue #429, opened 2016-06-13
Ability to inspect the primitive type used to store data


chadrik:
As a developer using schematics to automatically convert a model to an open schema format (e.g. json-schema), it would be very useful to be able to inspect the python data type used to store the primitive data (i.e. one of `str`, `int`, `float`, `bool`, `dict`, or `list`).  I propose a class attribute named `PRIMITIVE_TYPE`.  This is not something that would be required for a custom model to implement in order operate correctly, but it would be great for the built-in models to set the standard for users to follow.  

It can also be used to create variation through inheritance, as in:

``` python
class NumberType(BaseType):
    PRIMITIVE_TYPE = None
    ...

class IntType(NumberType):
    PRIMITIVE_TYPE = int
    ...

class FloatType(NumberType):
    PRIMITIVE_TYPE = float
    ...
```

bintoro:
+1. I've thought about this too.

`UnionType` probably shouldn't advertise this information at all. The main reason is that, in order to inspect the inner type of a compound type, the inspecting code would need to retrieve the underlying compound type instance anyway. So we might as well keep things simple and say:

* `PRIMITIVE_TYPE` is always a type (and not, say, a tuple of possibilities)
* `UnionType` fields should be examined by inspecting the underlying types directly.

chadrik:
Great. I'll make the pull request. Do you prefer caps for this variable?  There is some room for debate over whether abstract class attributes are static or not.

bintoro:
Hmm. This raises an important question: is the primitive type always static, or is it allowed to vary according to, say, a parameter on the field or the value being exported?

I was able to find one case where the type depends on the value: the new `TimestampType` returns an `int` if the value has a zero microsecond component; otherwise a `float` is returned. I think this can be classified as a bug — a `float` should always be returned. But is this kind of behavior always a bug?

Unless someone can provide a legitimate use case, I'd be inclined to rule that the primitive type cannot depend on context-specific circumstances such as the value itself or options specified at export time, as that would be incompatible with static schema descriptions.

As another example, consider a variant of `IPv4Type` that exports the value as an integer instead of a dot-decimal string (let's pretend it doesn't inherit from `StringType`). Would it be mandatory to implement this as a separate type instead of a parameter on `IPv4Type`? If not, then `PRIMITIVE_TYPE` may sometimes have to be a property and should not look like a constant.

[...]

PR #431, opened 2016-06-23
Allow primitive and native types to be inspected

chadrik:
This is a first pass at the feature discussed in issue #429

Some notes:
- instance variable `NumberType.number_class` has been replaced with class variable `NumberType.native_type`
- as discussed, `PolyModelType.native_type` is None because it can't be known
- as discussed, `TimestampType` always converts to float
- made a slight adjustment to ensure that `GeoPointType` always converts to `list`

Regarding `GeoPointType`, it might be nice for it to be a compound type, so that the data type of its components (`int`) can be inspected.  Also, it's a good demonstration of how schematics is meant to work.

I'm sure there will be some feedback, so let me know.

### DIFF
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -147,6 +147,9 @@ class BaseType(object):
           help, tooltips, documentation, etc.
     """
 
+    primitive_type = None
+    native_type = None 
+
     MESSAGES = {
         'required': "This field is required.",
         'choices': "Value must be one of {0}.",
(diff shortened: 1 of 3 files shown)


schematics/types/base.py
```diff
@@ -429,14 +438,14 @@ def _mock(self, context=None):
     def to_native(self, value, context=None):
         if isinstance(value, bool):
             value = int(value)
-        if isinstance(value, self.number_class):
+        if isinstance(value, self.primitive_type):
```
lkraider: Should probably use `self.native_type` in these tests and casts, no?

chadrik: `native_type` and `primitive_type` are the same, but I agree it would be more intuitive to use `primitive_type` within `to_primitive`.  If backward compatibility is a concern I could also keep `number_class`.

lkraider:
+1 code looks good to me.

lkraider:
Sorry for the delay, thanks for the patch!

PR #434, opened 2016-07-25
Support all non-string Iterables as choices

### LINKED ISSUE #433: Set is not a valid type for `choices` parameter
Using:

``` console
$ pip freeze | grep schematics
schematics==1.1.1
$ python --version
Python 2.7.10
```

And the following class:

``` python
from schematics.types import StringType

class Foobar(Model):
    foo = StringType(choices={'foo', 'bar'})
```

This raises a `TypeError`:

```
  File "test.py", line 4, in <module>
    class Foobar(Model):
  File "test.py", line 5, in Foobar
    foo = StringType(choices={'foo', 'bar'})
  File "/private/tmp/env/lib/python2.7/site-packages/schematics/types/base.py", line 339, in __init__
    super(StringType, self).__init__(**kwargs)
  File "/private/tmp/env/lib/python2.7/site-packages/schematics/types/base.py", line 184, in __init__
    raise TypeError('"choices" must be a list or tuple')
TypeError: "choices" must be a list or tuple
```

Is there a reason why `choices` must be an ordered collection?

Furthermore, this seems to be a regression, as this behavior was supported in `schematics<=1.0.2` and removed with https://github.com/schematics/schematics/pull/267

di:
Fixes #433, closes #436.

### DIFF
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -6,6 +6,7 @@ import itertools
 import functools
 import random
 import string
+from collections import Iterable
 
 import six
 from six import iteritems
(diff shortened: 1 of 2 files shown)


boblannon-picwell:
:+1:

PR #447, opened 2016-09-14
to_primitive should always return None when value is None

alexhayes:
If using a `DecimalType` and you supply a `None` value, when calling `to_primitive` it returns `'None'` (string).

This PR ensures that `None` is returned for all fields if the `value` of the field is `None`.

### DIFF
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -347,6 +347,8 @@ class UUIDType(BaseType):
         return value
 
     def to_primitive(self, value, context=None):
+        if value is None:
+            return None
         return str(value)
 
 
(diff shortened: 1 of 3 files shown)


lkraider:
Not sure about the use case here, when you serialize a Model the fields that have None will be in the output. This works today:

```python
def test_none_to_primitive():
    class Foo(Model):
        i = IntType()
        s = StringType()
        d = DecimalType()
        u = UUIDType()

    f = Foo()
    assert f.serialize() == {'i': None, 's': None, 'd': None, 'u': None}
```

lkraider:
Closing, but please open a new Issue or PR in case you think this is still needed.

PR #457, opened 2016-10-21
Import types explicitly for better readability

likel:
My IDE reports a error seems cause we don't import types explicitly. Maybe it's good to read if we import it manually. Happy to hear your comments. Thanks.


lkraider:
Added on development.

PR #462, opened 2016-11-27
Tweak code example to avoid sql injection

Ian-Foote:


lkraider:
Good point. Also fixes the query print output that didn't match the input string (by removing it).

Issue #446, opened 2016-09-13
Model + Schema


lkraider:
Hi all,

I am working on making latest Schematics work with SQLAlchemy (see https://github.com/schematics/schemalchemy/issues/1) by bringing calculated types (getter/setter functions) up-to-date.

The core issue is to have the model instance go through the import loop, so that custom user setters can run against the instance.

Towards this goal, I ended up splitting the Model into a simpler structure called Schema, similar to what SQLAlchemy does with its Table vs Base types.

The current work is ongoing in my repo at:
https://github.com/nKey/schematics/tree/calculated-type

What does this mean in practice? The Model now wraps the data and encodes the usual lib workflow, while the Schema encodes the actual type structure. This means any field name can fit into the Schema (no Model method conflict), and people that need more complex usage can rely on the backend without needing a Model.

Example:

``` python
    schema = Schema('Player',
        Field('id', IntType()),
        Field('first_name', StringType(required=True)),
        Field('last_name', StringType(required=True)),
        Field('full_name', calculated(
            type=StringType(),
            fget=get_full_name,
            fset=set_full_name))
    )
    input_data = {'id': '1', 'full_name': 'John Doe', 'hi': 1}
    converted_data = convert(schema, input_data, partial=True)
    validated_data = validate(schema, converted_data, convert=False, partial=True)
```

This issue is just a notice I am working on this, it is not ready for a PR and it breaks the API (although I am making compatibility mixins).

PS: any and all feedback is welcome

lkraider:
I've implemented most of the Schema, changed the Model dict structure and updated most of the code to pass the tests.

There is some work yet to be done in the Compound types and then porting to work with all different Python versions.

For some notable changes, see:
https://github.com/nKey/schematics/blob/calculated-type/schematics/schema.py
https://github.com/nKey/schematics/blob/calculated-type/tests/test_schema.py
https://github.com/nKey/schematics/blob/6140d3ff189cd8cfeb78e0ea5d1e2f3ed4766904/schematics/models.py#L143

lkraider:
All tests are passing, work now is to test with SQLAlchemy integration.

PR #466, opened 2016-12-09
Schema API

lkraider:
Implements #446 

This is the foundational work to get back compatibility with SQLAlchemy. To do that, the concept of Serializable setters is reintroduced, along with a lower level abstraction of the Model named Schema.

The functional API can now rely on the Schema declaration directly, which simplifies integration by exposing a smaller surface. More work will need to be done to introduce a CompoundType with pure Schema.

Additionally, the concept of data validation stages is introduced (see `contrib/machine.py`) and initial work has been done in the Model to apply that, but it's not fully realized yet. Discussion for a more granular approach will be presented in another PR.

### DIFF
--- a/schematics/__init__.py
+++ b/schematics/__init__.py
@@ -2,6 +2,10 @@
 
 __version__ = '2.0.0.a1'
 
+# TODO: remove deprecated API
+from . import deprecated
+deprecated.patch_all()
+
 from .models import Model, ModelMeta
 
 types.compound.Model = Model
--- a/schematics/__init__.py
+++ b/schematics/__init__.py
@@ -6,6 +6,7 @@ __version__ = '2.0.0.a1'
 from . import deprecated
 deprecated.patch_all()
 
+from . import types
 from .models import Model, ModelMeta
 
 types.compound.Model = Model
--- a/schematics/compat.py
+++ b/schematics/compat.py
@@ -31,8 +31,9 @@ else:
 def metaclass(metaclass):
     def make_class(cls):
         attrs = cls.__dict__.copy()
-        del attrs['__dict__']
-        del attrs['__weakref__']
+        if attrs.get('__dict__'):
+            del attrs['__dict__']
+            del attrs['__weakref__']
         return metaclass(cls.__name__, cls.__bases__, attrs)
     return make_class
 
--- a/schematics/compat.py
+++ b/schematics/compat.py
@@ -23,9 +23,11 @@ if PY2:
     from itertools import imap as map
     from itertools import izip as zip
     iteritems = operator.methodcaller('iteritems')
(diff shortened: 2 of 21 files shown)


lkraider:
This is ready for merging. Fixes SQLAlchemy integration.

I am leaving the `deprecated` API compatibility-layer enabled by default, work will need to be done to refactor it out later.

Unless there are objections, I'll merge it this friday.

bintoro:
@lkraider Could you maybe illustrate what's going on here in an easy-to-understand fashion, if possible? :)

lkraider:
@bintoro The motivation for this work was to bring serializable setters (from 0.9) back into mainline. This feature is required for SQLAlchemy integration (supported by https://github.com/schematics/schemalchemy).

To make this possible it was required to change the transform functions back to accept and operate on instance data, as setters in the model have to modify the instance state (this also allows instance validators again).

My approach was to avoid making the Model instance a requirement, but to have them accept a "mutable" mapping, so that non-model data can be validated too. To make this explicit I opted to separate the Model from the actual "Schema" specification that is really required during the transforms. I took inspiration in the architecture in SQLAlchemy that splits the Table description from the `declarative` Base model mapping. This split helps in that it separates concerns of the structure itself from its workflow (the Schema encodes the structure, the Model encodes the workflow).

Related to the setters introduction, means that they can generate "raw data" during a transform (separated into its own `_mutate` method), but for the Model means there is a need to split valid/raw data, so now the Model dict has that segregation builtin. There is another way to achieve that which I want to explore further, and that is to attach metadata to the values to indicate their state, but that requires more discussion.

For this PR, I kept the Schema/Model interchangeable by providing some "deprecated" API mixins. The intent is to drop them completely eventually.

Below is an overview of the encoded states now:
schematics-state-machine

As you can see, each step will only operate with data that is in the correct state from the previous step (at least that is the idea), and mutation resets the data state.

bintoro:
At this point I barely understand any of the diffs. This is going to take a long time to review. Would someone else like to give it a shot?

lkraider:
Sorry for the long delay, I am ready to merge it this week.

bintoro:
First question: why not have `Model` inherit from `Schema`?

lkraider:
I preferred composition in this case, to make the split more clear. Also because the Model is already complicated enough with metaclass behaviour and properties, I opted not to overload it with more attributes (that could conflict with user-defined properties). It also makes the interaction with SQLAlchemy Base easier to understand.

The Model responsibilities now is to encapsulate API usage, the schema and data state.

lkraider:
I will move forward with this patch into develop branch this weekend, it should not break high level compatibility with current 2.0.0a code. People can propose changes with other PRs.

Issue #452, opened 2016-09-29
ValidationError is not Hashable (2.0.0)


Guibod:
I can't `logging.exception(ValidationError)` it seems that the object is not hashable.

That's sad, it used to work pretty well on 1.1. I'm using Python 3.4.

```
class TestUnhashable(unittest.TestCase):
    def test_type_error_is_hashable(self):
        try:
            raise TypeError('Foo bar')
        except TypeError as e:
            logging.exception(e)

    def test_validation_error_is_unhashable(self):
        try:
            raise ValidationError('Foo bar')
        except ValidationError as e:
            with self.assertRaises(TypeError) as c:
                logging.exception(e)

            self.assertIsInstance(c.exception, TypeError)
            self.assertEqual(c.exception.args[0], 'unhashable type: \'ValidationError\'')
```

Guibod:
Correction. I have no clue if it was working in v1.1. Yet this would be great to be able to `logging.exception(ValidationError)`

lkraider:
Your code works in Python 2, but breaks in Python 3 because it auto-sets `__hash__ = None` when `__eq__` is defined for a class (as is the case in Schematics `ErrorMessage`) - you need to provide your own hash method.

In Python an object can only be [hashable](https://docs.python.org/3/glossary.html#term-hashable) if it has a hash value which never changes during its lifetime, that is, it is immutable.

We have a "problem" right now with the Errors in that they are mutable (they contain a list of messages and a `pop` method that changes it).

Also, hashable objects which compare equal must have the same hash value, and we use the instance `__dict__` for comparison inside `__eq__`, which is also mutable/unhashable.

Python itself does not require Exceptions to be immutable. At least in 2002, GVR was in favor of mutable exceptions: https://mail.python.org/pipermail/python-dev/2002-August/027900.html

The `logging` framework assumes exceptions are hashable (and thus immutable), which is an undocumented assumption. But since logging is in the stdlib, we need to play by their rules.

The fix to support logging is to make all schematics exceptions immutable. I don't have a proposal for how to do this, but would welcome a PR.

lkraider:
Fixed in #477

Issue #445, opened 2016-09-06
Parsing exception messages in 2.0.0a1


synic:
In 1.x, validation error exceptions could be returned directly to the client.  Now they are in the format:

```
DataError({'email': [ErrorMessage("A calendar with the email address "jasonmartinez@nichols.org" already exists in the system.")]},)
```

If there's a `ModelType` field on the model, the error messages can even come with embedded `ValidationError` classes in it, like so:

```
{'credentials': {'oauth2_credentials': ValidationError("Invalid credentials supplied.")}}
```

Are there any current methods for making these easier to deal with?  How is it expected we'll convert these error messages?

neilalbrock:
I'd also like some clarification on this... I've previously used schematics to model and validate API calls and the 1.x handling of errors made it simple to return those in JSON format to the consumer on error.

In 2.x I'm having to write some fairly byzantine code to handle getting from the new error format to something that can be serialised to JSON. This is a retrograde step in my opinion, so at least providing a new equivalent to the old errors block would simplify things again.

Schematics makes doing fairly complex things nice and simple, be good to try and keep it that way.

rmb938:
I would like to easily convert these errors to JSON as well.

synic:
I worked around it by creating a custom JSONDecoder that basically calls `str` on anything it doesn't recognize:

```python
import json

class ExceptionJSONEncoder(json.JSONEncoder):
    """Format the error messages we get from schematics.
    I think the format of their error messages is currently bugged, I reported
    it here: https://github.com/schematics/schematics/issues/445
    Hopefully this class is just temporary.
    """
    def default(self, obj):
        return str(obj)
```

Then you can do something like `json.dumps(data, cls=ExceptionJSONEncoder)`.  It works, but it's definitely not ideal.

neilalbrock:
I ended up with this:

```python
def serialize_errors(errors):
    """Serialize Schematics 2.x Errors"""
    if isinstance(errors, dict):
        res = {}

        for k,v in errors.iteritems():
            if isinstance(v, basestring):
                res[k] = v
            elif isinstance(v, list):
                res[k] = [m.summary for m in v]
            elif isinstance(v, dict):
                res[k] = serialize_errors(v)
            else:
                res[k] = [m.summary for m in v.messages]

        return res

    else:
        return errors
```

I'll give the JSON encoder version a try, might be simpler. Although I wonder if this version might allow for serializing to XML too?

bintoro:
[...]

Issue #369, opened 2015-11-28
Redesign the conversion/validation exceptions


bintoro:
### Unify aggregate exception classes

The current separation between `ModelConversionError` and `ModelValidationError` provides no informational value at all, since the error class is picked solely based on which model-level routine was invoked and has nothing to do with what actually went wrong.

To illustrate, the _exact same_ conversion error may appear as a `ModelConversionError` or `ModelValidationError`, depending on if you called `Model.convert()` or `Model.validate()`, respectively. This is unnecessary complexity and a potential source of confusion (#137).

Also, since these classes may represent data on _several_ errors, each of which may independently be either a conversion error or a validation error, it's clear that communicating anything about the error types through the name of the class is just not viable.

Based on this, I think the aggregate exception classes should be unified.
### Improve exception message format

Right now the aggregate error message structure only contains the names of the offending fields coupled with textual messages.

The format should be redesigned to accommodate at least the following items for each individual error:
- name of the field or index of the item (#307), whichever is applicable
- type of the error (conversion/validation)
- human-readable textual message
- structured response data from the validator/converter function (e.g., error code).

bintoro:
I've been sketching a new exception hierarchy that looks like this:

* BaseError
  * FieldError
    * ConversionError
    * ValidationError
    * StopValidationError
  * CompoundError
    * DataError

All `FieldError` subclasses represent field-specific errors arising from the field's own validators, while errors bubbling up from substructures within compound fields are represented by a `CompoundError`.

Therefore, `FieldError.messages` is now guaranteed to be a flat list (as is currently the case for errors on basic types).

`CompoundError.messages`, on the other hand, is always guaranteed to be a dictionary, keyed by the offending item's identifier. For a `ListType`, this identifier is the item's index.

`DataError` is a `CompoundError` representing the result of validating an entire model. It's basically identical to `ModelConversionError` and `ModelValidationError`, both of which will be deprecated.

Error messages are no longer just strings but `ErrorMessage` objects containing three attributes:
  * type (conversion or validation error)
  * summary (the human-readable message)
  * info (an arbitrary value; e.g., an error code or some other response from the validator)

#### Example

```python
class M(Model):
    intfield = IntType(max_value=3)
    listfield = ListType(IntType(max_value=3))
    modelfield = ModelType('M')

inputdict = {
    'intfield': 1,
    'modelfield': {
        'intfield': 2,
        'modelfield': {
            'intfield': 3,
[...]

Issue #465, opened 2016-12-08
Unmaintained


rmb938:
Is it safe to assume this project is no longer maintained? Considering there has been very little activity since May of early this year. 

@bintoro @jmsdnns @kracekumar If you guys need more people to help out I am more than happy to lend a hand.

I use schematics in a lot of my projects since there are not any other libraries that have similar features. I would like to continue using it but I am currently forced to make bug fixes and add features on my own forks.

bintoro:
The situation sucks, but please do not despair. I've been swamped with actual income-generating work, but I fully intend to get back to Schematics with a vengeance.

@rmb938, would you like to submit issues/PRs for some of the bugs you speak of?

rmb938:
Well most of these already have prs open including my own; #449, #447, #436. As well as some sort of fix/improvement for #445

lkraider:
I am working in schematics in my own fork to fix the SQLAlchemy integration (and more), it is getting ready to merge here.

I guess I could also work directly in this repo so it doesn't look dead...

tuukkamustonen:
Overall, I don't think that it's optimal situation that we have (at least) 3 competing validation/serialization frameworks out there: `schematics`, `marshmallow` and `voluptuous`. Yeah, each of them has a slightly different focus, but as a silly customer I would love to see the power of all in just one package.

At the moment I'm using marshmallow for parameter validation (for it has built-in integration via `webargs`), voluptuous in tests (doing some asserts with it because it has the least-boilerplate syntax) and schematics for remaining code (compared to marshmallow, I access the data via `Model` instance, no need for custom construction). I think first schematics, then discovered voluptuous and then marhsmallow/webargs, which left me using different libraries for different tasks... yeah, one could say that's the way to go ("use the right tool for the task") but I don't quite agree.

If we had one strong/powerful/flexible package that would get the love of all the authors, that would rock. There's lots of duplication over the projects. But, I know I'm whining here, just wanted to speak out my dreams :)

harlowja:
Agreed, all seem to do similar stuff; each a little bit differently.

kracekumar:
@rmb938 Glad that you brought this up. In my opinion, schematics is stable and has required API's to work for most of the cases.

I haven't been contributing to the project since there isn't any active development other than @bintoro's work. He knows what he is doing.At my last job, I used `schematics` as validation library for Flask API. So feel free to get involved. I am all up for new contributors who can make schematics better.

[...]

PR #477, opened 2017-03-18
Enhancing and addressing some issues around exceptions:

toumorokoshi:
This PR attempts to address a few issues around exceptions:

#452: all exceptions are now hashable and immutable
#445: exceptions now have a human readable message, and a to_primitive function is provided to for serialization in markup (e.g. json)
#369: conforms to PEP-352, and normalizes all exceptions to a message and an errors field.

This PR has a backward-incompatible change of FieldErrors no longer being mutable, which I'm amenable to. FieldError.pop() is removed as a result.

### DIFF
--- a/schematics/exceptions.py
+++ b/schematics/exceptions.py
@@ -77,7 +77,7 @@ class BaseError(Exception):
         return json.dumps(self.to_primitive())
 
     def __repr__(self):
-        return "<%s: %s>" % (self.__class__.__name__, repr(self.errors))
+        return "%s(%s)" % (self.__class__.__name__, repr(self.errors))
 
     def __hash__(self):
         return hash(self.errors)
@@ -102,7 +102,11 @@ class ErrorMessage(object):
         self.info = info
 
     def __repr__(self):
-        return '%s(%s)' % (self.__class__.__name__, str(self))
+        return '%s(%s, %s)' % (
+            self.__class__.__name__,
+            repr(self.summary),
+            repr(self.info)
+        )
 
     def __str__(self):
         if self.info:
--- a/schematics/exceptions.py
+++ b/schematics/exceptions.py
@@ -2,9 +2,73 @@
 
 from __future__ import unicode_literals, absolute_import
 
+import copy
+
 from collections import Sequence
 
 from .common import *
+
+
+@str_compat
+class BaseError(Exception):
+
+    def __init__(self, message, errors=None, *args):
+        """
+        The base class for all Schematics errors.
+
+        message should be a human-readable message,
+        while errors is a machine-readable list, or dictionary.
+
+        if None is passed as the message, and error is populated,
+        the primitive representation will be serialized.
+
+        the Python logging module expects exceptions to be hashable
+        and therefore immutable. As a result, it is not possible to
+        mutate BaseError's error list or dict after initialization.
+        """
+        if message is None and errors is not None:
+            message = str(self._to_primitive(errors))
+        super(BaseError, self).__init__(message)
+        self._errors = errors
+
+    @property
+    def errors(self):
+        return copy.deepcopy(self._errors)
+
+    @property
+    def messages(self):
+        """ an alias for errors, provided for compatibility with V1. """
+        return self.errors
+
+    def to_primitive(self):
+        """
+        converts the errors dict to a primitive representation of dicts,
+        list and strings.
+        """
+        return self._to_primitive(self._errors)
+
+    @classmethod
+    def _to_primitive(cls, obj):
+        """ recursive to_primitive for basic data types. """
+        if isinstance(obj, list):
+            return [cls._to_primitive(e) for e in obj]
+        elif isinstance(obj, dict):
+            return dict(
+                (k, cls._to_primitive(v)) for k, v in obj.items()
+            )
+        else:
+            return str(obj)
+
+    def __hash__(self):
+        if not hasattr(self, "_hash"):
+            if isinstance(self._errors, list):
+                self._hash = hash(tuple(self._errors))
+            if isinstance(self._errors, dict):
+                dict_as_tuple = tuple([
+                    (k, self._errors[k]) for k in sorted(self._errors)
+                ])
+                self._hash = hash(dict_as_tuple)
+        return self._hash
 
 
 @str_compat
@@ -40,9 +104,8 @@ class ErrorMessage(object):
         else:
             return False
 
-
-class BaseError(Exception):
-    pass
+    def __hash__(self):
+        return hash((self.summary, self.type, self.info))
 
 
 @str_compat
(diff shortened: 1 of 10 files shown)


toumorokoshi:
@lkraider here's a PR to unblock #465

lkraider:
Thanks for the PR and addressing the mutability as well, I'll review it today!

schematics/exceptions.py
```diff
@@ -2,12 +2,76 @@
...
+        mutate BaseError's error list or dict after initialization.
+        """
+        if message is None and errors is not None:
+            message = str(self._to_primitive(errors))
+        super(BaseError, self).__init__(message)
+        self._errors = errors
+
+    @property
+    def errors(self):
+        return copy.deepcopy(self._errors)
```
lkraider: **Improvement**: The deepcopy is safe but could be an expensive side-effect for a property to wrap, can't we use an immutable primitive for the errors? A tuple of namedtuples, or something like https://github.com/slezica/python-frozendict ?

schematics/exceptions.py
```diff
@@ -66,55 +129,55 @@ def __init__(self, *args, **kwargs):
...
-        if len(self.messages) == 1:
-            msg_repr = str(self.messages[0])
+        if len(self.errors) == 1:
+            msg_repr = str(self.errors[0])
         else:
-            msg_repr = '[' + str.join(', ', map(str, self.messages)) + ']'
+            msg_repr = '[' + str.join(', ', map(str, self.errors)) + ']'
         return '%s(%s)' % (self.__class__.__name__, msg_repr)
 
     __str__ = __repr__
```
lkraider: We need to solve the quoting issue and probably return a JSON parseable string:

```python
try:
    Player({'id': 1.2, 'location': [2]})
except DataError as e:
    print(str(e))
```

Now:
```
{'id': u'ConversionError("Value \'1.2\' is not int.")', 'location': u'ConversionError("Input must be a mapping or \'Location\' instance")'}
```

Desired:
```
{"id": "Value '1.2' is not int.", "location": "Input must be a mapping or 'Location' instance"}
```

schematics/exceptions.py
toumorokoshi: I'd vote for frozendict, the dict representation is nice without having to serialize to / from tuples. 

Would you like me to pull frozendict as a dependency, or write one for schematics?

schematics/exceptions.py
toumorokoshi: sounds good, I can make the string more legible. 

I feel like it shouldn't be guaranteed behavior that the string cast is json parseable: it's a strange restriction to put on a call that's supposed to be a human-readable representation of the errors.

I'd argue for documenting the use of something like to_primitive() for those who want to serialize output in whatever format they desire (e.g. json or yaml)

schematics/exceptions.py
lkraider: I think we can pull a copy of frozendict into datastructures.py, it's simple enough.

schematics/exceptions.py
lkraider: Indeed, the use case is to be human-readable. I was thinking of the json structure for the simplicity of serializing the dict into a `field_name: message` string, what structure would you propose?

toumorokoshi:
Here's an updated version, addressing some of the PR feedback:

1. string representations of error messages are equivalent to json formatting a primitive representation. Agreed json is a pretty legible format, just want to note it's an implementation detail to serialize to json, and shouldn't be relied on. (serializing to_primitive exists for that purpose)

1a. this also involved modifying the string formatting around error message a bit, so it doesn't write out as tuples in the case there's a summary and info included: it was clunky to put a bunch of escaped quotes in the values, if we're targeting human readability.

2. dicts and lists are converted to FrozenDict and FrozenList respectively. to note: I wrote my own implementation of FrozenDict to avoid pulling in the MIT license of the FrozenDict code. modified unit tests that relied on mutable lists.

3. I spent some time examining the FieldError code, and was able to pull more of that checking into the BaseError. This ensures that all the conveniences of FieldError (e.g. __eq__ of the exception with the underying list or dict) is now consistent for all BaseError types.

schematics/exceptions.py
```diff
@@ -26,49 +29,71 @@ def __init__(self, message, errors=None, *args):
...
         converts the errors dict to a primitive representation of dicts,
         list and strings.
         """
-        return self._to_primitive(self._errors)
+        if not hasattr(self, "_primitive"):
+            self._primitive = self._to_primitive(self._errors)
+        return self._primitive
+
+    @staticmethod
+    def _freeze(obj):
```
toumorokoshi: I wanted to note that it's still possible to modify elements and dictionary values of FrozenDict and FrozenList. At a performance cost, I can do a recursive _freeze to guard against this.

lkraider: I think we can keep it so, it's pretty hard to enforce complete immutability in Python anyways, and making the base object immutable is a good enough indication (for consenting adults) that it shouldn't be changed.

schematics/exceptions.py
```diff
@@ -2,9 +2,98 @@
...
+            return dict(
+                (k, cls._to_primitive(v)) for k, v in obj.items()
+            )
+        else:
+            return str(obj)
+
+    def __repr__(self):
+        return json.dumps(self.to_primitive())
+
+    __str__ = __repr__
```
lkraider: Do we need `repr` to be equal `str`? We can have repr return a "reconstructable" string, as that is its recommended format.

tests/test_exceptions.py
```diff
@@ -86,7 +87,7 @@ def test_error_eq():
     assert ValidationError("A") != ValidationError("A", "B")
 
 
-def test_error_pop():
+def _test_error_pop():
```
lkraider: Let's remove this test, this functionality will not be supported.

schematics/exceptions.py
```diff
@@ -2,9 +2,98 @@
 
 from __future__ import unicode_literals, absolute_import
 
-from collections import Sequence
+import copy
```
lkraider: Unused import of `copy`.

toumorokoshi:
@lkraider addressing the feedback.

I was able to make repr work with the eval, as specified.

I also noticed something about PEP 352: the only reference to using a string as the first argument is for the special casing that is included in the BaseException class: it looks like the type of the first argument is not so important.

even the examples of user-defined exceptions violates the message as the first argument:

https://docs.python.org/3/tutorial/errors.html

As such, I modified the behavior to not take a string, and to always take a list or dict object as the first argument. This gets rid of the clumsy user-driven string, which was not used anywhere.

bintoro:
Once this work is complete, could one of you squash the commits together, please?

toumorokoshi:
I would squash, but I believe that erases comment history. I'll leave it up to the merger unless it's ok for me to overwrite history here. 

The repr check has an issue with the unicode encoding and Python2. I'll try to tackle it tomorrow.

lkraider:
@toumorokoshi The code is looking great!

@bintoro sure thing, I'll squash&merge from the github interface when it's done.

toumorokoshi:
@lkraider so in tackling that repr issue, I found another fun issue. I've resolved it to a pretty good point, but I wanted to make note of it.

the str_compat decorator on classes was interfering with the ability to make nested repr calls. In the case of CompoundError in Python2, the ErrorMessage class would be converted to a unicode-escaped string, which would then be double-escaped when the BaseError escapes ITS __repr__ call.

Removing str_compat form BaseError and ErrorMessage resolves the issue, but I imagine it loses the __str__ and __unicode__ behavior that was expected in the first place. Thus, I split the str_compat into str_compat and repr_compat. I ran into a weird recursive import with attempting to use repr_compat on the model. I put in a quick workaround, but I'll take a closer look if I can. 

Ultimately, the rational for operating on __str__ / __unicode__ makes sense to me, but attempting to escape __repr__ makes nested repr calls impossible, and it's unnecessary as long as __repr__ overrides compose their parts out of other __repr__ calls (such as how BaseError / ErrorMessage now works).

At this point this addresses all the feedback, and maintains as much compatibility as possible (the broken places were already mentioned above, like mutability).

it might be fun to tackle the repr_compat issue across all of schematics in a separate PR. But if desired I can do it in this one.

toumorokoshi:
Ok, I think this is good to go now. I removed the hacky with_repr in str_compat (it was an ImportError hidden under a few other messages), and fixed a flakey unit test.

lkraider:
@toumorokoshi Thank you for the great patch, squashed and applied to development!

toumorokoshi:
Thanks! Great feedback, pleasure to work with you, and looking forward to the 2.0 release!

PR #487, opened 2017-05-13
Add validation using role

aleksej-paschenko:
Hi there,
In our project we use single model with 2 roles like in this snippet:

```python
    class Address(Model):
        class Options:
            roles = {
                'full': whitelist('street', 'zipcode'),
                'short': whitelist('street')
            }
        zipcode = StringType(required=True)
        street = StringType(required=True)

    class User(Model):
        class Options:
            roles = {
                'full': whitelist('name', 'age', 'address'),
                'short': whitelist('name', 'address')
            }
        name = StringType(required=True)
        age = IntType(required=True)
        address = ModelType(Address, required=True)
```
And I can not understand why there is no validation using some role. 
When role=short I think that validation should be done without exceptions for such json: 
```javascript
 {
        'name': 'Bob',
        'address': {
            'street': 'somewhere in the world'
        }
}
```
Can somebody point me in the right direction If there are any problems with PR

### DIFF
--- a/schematics/models.py
+++ b/schematics/models.py
@@ -225,7 +225,7 @@ class Model(object):
         if validate:
             self.validate()
 
-    def validate(self, partial=False, convert=True, app_data=None, **kwargs):
+    def validate(self, partial=False, convert=True, app_data=None, role=None, **kwargs):
         """
         Validates the state of the model. If the data is invalid, raises a ``DataError``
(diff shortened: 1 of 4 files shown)


lkraider:
A role is used only for serialization, not for validation. You can use Model inheritance to describe the extended fields.

PR #488, opened 2017-05-16
Add type hints to constructors of field type classes

KonishchevDmitry:
I've added type hints to constructors of field type classes to suggest IDE the actual type of the object and now PyCharm deduces types of all fields in model objects, even for the nested ones like `model.model_list[0].nested_string` which make auto-completion and refactoring to work much better.

See the screenshot:
Field type hinting in PyCharm

Code I've tested it on:

```python
from typing import List, Dict, Union, Tuple

import datetime
from decimal import Decimal

from schematics.types import StringType, ListType, ModelType, NumberType, IntType, FloatType, DecimalType, \
    BooleanType, DateType, DateTimeType, DictType

class TestNestedModel(Model):
    nested_string = StringType()

class TestModel(Model):
    string = StringType()
    boolean = BooleanType()
    integer = IntType()
    float = FloatType()
    number = NumberType()
    decimal = DecimalType()
    date = DateType()
    time = DateTimeType()
    dict = DictType(TestNestedModel)
    string_list = ListType(StringType())
    model = ModelType(TestNestedModel)
    model_list = ListType(ModelType(TestNestedModel))

class Bad: pass
invalid_type = Tuple[Bad, Bad, Bad, Bad, Bad, Bad, Bad, Bad, Bad, Bad, Bad, Bad, Bad]
valid_type = Tuple[str, bool, int, float, Union[int, float], Decimal, datetime.date, datetime.datetime,
                   Dict[str, TestNestedModel], List[str], TestNestedModel, List[TestNestedModel], str]

def test(valid: valid_type, invalid: invalid_type):
    pass

model = TestModel()
test_data = (
    model.string, model.boolean, model.integer, model.float, model.number, model.decimal, model.date,
    model.time, model.dict, model.string_list, model.model, model.model_list, model.model_list[0].nested_string
)
test(test_data, test_data)
```

### DIFF
--- a/schematics/types/compound.py
+++ b/schematics/types/compound.py
@@ -2,6 +2,13 @@
 
 from __future__ import unicode_literals, absolute_import
 
+try:
+    import typing
+except ImportError:
+    pass
+else:
+    T = typing.TypeVar("T")
+
 from collections import Iterable, Sequence, Mapping
 import itertools
 
@@ -84,7 +91,10 @@ class ModelType(CompoundType):
     def fields(self):
         return self.model_class.fields
 
-    def __init__(self, model_spec, **kwargs):
+    def __init__(self,
+                 model_spec,  # type: typing.Type[T]
+                 **kwargs):
+        # type: (...) -> T
 
         if isinstance(model_spec, ModelMeta):
             self.model_class = model_spec
@@ -156,7 +166,11 @@ class ListType(CompoundType):
     primitive_type = list
     native_type = list
 
-    def __init__(self, field, min_size=None, max_size=None, **kwargs):
+    def __init__(self,
+                 field,  # type: T
+                 min_size=None, max_size=None, **kwargs):
+        # type: (...) -> typing.List[T]
+
         self.field = self._init_field(field, kwargs)
         self.min_size = min_size
         self.max_size = max_size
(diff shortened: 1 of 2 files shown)


lkraider:
That's really cool. I was playing with the typing module myself for the internals, I think it merits to be included.

lkraider:
Hi, can you rebase and remove these comments:
```python
# ^ Override return type to suggest IDE the actual type of the object
```

I'll then merge it in!

schematics/types/compound.py
```diff
@@ -83,7 +90,11 @@ def native_type(self):
     def fields(self):
         return self.model_class.fields
 
-    def __init__(self, model_spec, **kwargs):
+    def __init__(self,
+                 model_spec,  # type: typing.Type[T]
```
lkraider: Can't this be represented in the doctype?

KonishchevDmitry: It seems that it can't. I've tried various forms like `# type: (typing.Type[T], ...) -> T`: it works for the `model_spec` argument, but breaks other arguments (`**kwargs`, and named parameters).

KonishchevDmitry:
OK, removed. I added these comments just to prevent some contributors who don't know about type hinting from accidental break of my changes.

lkraider:
Thanks for the cool patch!

KonishchevDmitry:
Thank you!

PR #489, opened 2017-05-23
dictionary-named methods are raising a validation exception

toumorokoshi:
There's currently a bug where naming fields of a model after the dictionary methods will raise an incorrect DataError during validation (see unit test or https://github.com/toumorokoshi/transmute-core/issues/20)

Looking at the documentation, the atoms() method shouldn't support anything besides a mapping anyway. This builds off of that assumption, and removes the getattr call.

### DIFF
--- a/schematics/iteration.py
+++ b/schematics/iteration.py
@@ -1,4 +1,3 @@
-
 from .compat import iteritems
 from .undefined import Undefined
 from collections import namedtuple
@@ -44,9 +43,7 @@ def atoms(schema, mapping, keys=Atom._fields, filter=None):
 
         if has_value:
             try:
-                value = getattr(mapping, field_name)
-            except AttributeError:
-                value = mapping.get(field_name, Undefined)
+                value = mapping[field_name]
             except Exception:
                 value = Undefined
 
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -10,6 +10,50 @@ from schematics.types import StringType, IntType, ListType, ModelType
 from schematics.exceptions import *
 
 
+def test_dict_methods_in_model():
+    """
+    a regression test to ensure that an issue where attributes on
+    dictionaries are not being misintrepreted as actual schematics
+    fields.
+    """
+
+    class M(Model):
+        items, values, get = IntType(), IntType(), IntType()
+
+    m = M({"items": 1, "values": 1, "get": 1})
+    m.validate()
+
+
+def test_dict_methods_in_model_atoms():
+    """
+    atoms should return the raw values, and not call any overriden methods.
+    """
+    class M(Model):
+        get = IntType()
+    m = M({"get": 1})
+    atom = list(m.atoms())[0]
+    assert atom.name == "get"
+    assert atom.value == 1
+
+
+def test_nested_model_override_mapping_methods():
+    """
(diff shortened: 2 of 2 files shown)


lkraider:
Indeed this will fix this test case, but users that override other mapping methods (ex: "keys") will still break (relevant: #375). Also, overriding "get" will break other things (ex: `m.atoms()` will not work anymore to retrieve data), so you might get `Undefined` values depending on the case, as it will break this current assumption:
```
convert(m._schema, m) == convert(m._schema, m._data)
```

I could go ahead and merge this, but not sure a new release is warranted until a more general fix is achieved.

toumorokoshi:
@lkraider thanks, that ticket provided a lot of context. I see why this doesn't solve the larger collisions that occur between field names and the reserved method names of the Mapping type itself.

It sounds like there' s a lot of work to get to the point of separating the two apart (and a significant part backwards-incompatible). I can avoid dealing with this head on by using serialize_name and deserialize_from for now.

I would still argue this patch should be considered. It's a minimal change that allows better compatibility with the set, at least as far as the core schematics serialize logic is concerned. I actually think embarking on the larger solution is a good long-term goal, but I personally don't have the bandwidth to tackle that right now. 

I've made a change to rely on __getitem__ instead of get. This ensures that overriding any public method will not break schematics ability to serialize / deserialize.

schematics/iteration.py
```diff
@@ -1,4 +1,3 @@
-
```
lkraider: Please don't change whitespace formatting in this patch.

tests/test_models.py
```diff
@@ -658,4 +684,3 @@ class M(Model):
         m = ListType(ModelType('M', required=True), required=True)
 
     M.get_mock_object()
-
```
lkraider: See first comment.

tests/test_models.py
```diff
@@ -10,6 +10,32 @@
...
+    dictionaries are not being misintrepreted as actual schematics
+    fields.
+    """
+
+    class M(Model):
+        items, values, get = IntType(), IntType(), IntType()
+
+    m = M({"items": 1, "values": 1, "get": 1})
+    m.validate()
+
```
lkraider: Please add a test for nested ModelType that override mapping methods too:

```python
def test_nested_model_override_mapping_methods():

    class Nested(Model):
        items, values, get = IntType(), IntType(), IntType()

    class Root(Model):
        values = ModelType(Nested)

    root = Root({"values": {"items": 1, "values": 1, "get": 1}})
    root.validate()
    assert root.values.items == 1
```

lkraider:
I put forward a patch on master that fixes the reliance on `keys` iteration. You can amend the test-cases to include a keys attribute.

toumorokoshi:
Sounds good! Addressed all feedback, I believe. Thanks for fixing up the keys behavior as well!

Also just rebased on top of your changes, hope that's ok. not sure how that affects the existing CR.

lkraider:
Let me know if you have issues with the latest master. I will prepare a 2.0.1 release including this fixes later in the week.

lkraider:
FYI: v2.0.1 just released to PyPI and contains these fixes.

toumorokoshi:
works for me, thanks!

PR #491, opened 2017-05-28
DateTimeType to_native method should handle type errors gracefully

e271828-:


lkraider:
Thanks for the fix!

e271828-:
My pleasure. We ran into this one by chance. I suspect there may be similar
issues in other types, but have not done an audit.

lkraider:
FYI: v2.0.1 just released with these changes.

PR #495, opened 2017-06-08
Add possibility to translate strings

jaysonsantos:
To integrate with flask_babel it is simple as doing this:

```python
from flask_babel import lazy_gettext
from schematics.translator import register_translator
register_translator(lazy_gettext)
```

### DIFF
--- a/schematics/translator.py
+++ b/schematics/translator.py
@@ -1,6 +1,8 @@
 # coding: utf-8
+from .compat import str_compat
 
 
+@str_compat
 class LazyText(object):
     def __init__(self, message):
         self.message = message
--- a/schematics/translator.py
+++ b/schematics/translator.py
@@ -1,0 +1,22 @@
+class LazyText(object):
+    def __init__(self, message):
+        self.message = message
+
+    def __str__(self):
+        translator = _.real_translator
+        return translator(self.message) if translator else self.message
+
+
+class Translator(object):
+    """A placeholder which could call a function like lazy_gettext and make messages translatable."""
+    def __init__(self):
+        self.real_translator = None
+
+    def __call__(self, message, lazy=True, *args, **kwargs):
+        return LazyText(message) if lazy else str(LazyText(message))
+
+    def register_translator(self, new_translator):
+        self.real_translator = new_translator
+
+_ = Translator()
+register_translator = _.register_translator
--- a/schematics/translator.py
+++ b/schematics/translator.py
@@ -1,3 +1,6 @@
+# coding: utf-8
+
+
 class LazyText(object):
     def __init__(self, message):
         self.message = message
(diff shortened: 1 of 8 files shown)


lkraider:
Thanks for the patch, we've been missing i18n support for a while (had a patch on 0.9 but never integrated). I'll review this and merge when approved.

jaysonsantos:
The way I did is not enforce any tool. We use flask_babel here and to extract the strings we just use `pybabel extract schematics_folder` and translated by ourselves. I *think* Transifex has a free option for open source projects where community could help translating the project.

schematics/translator.py
```diff
@@ -0,0 +1,22 @@
+class LazyText(object):
```
lkraider: Please add these lines to make the class unicode-compat aware:
```python
# -*- coding: utf-8 -*-

from .compat import *


@str_compat
class LazyText(object):
```

schematics/types/base.py
```diff
@@ -321,7 +322,7 @@ def check_required(self, value, context):
     def validate_choices(self, value, context):
         if self.choices is not None:
             if value not in self.choices:
-                raise ValidationError(self.messages['choices']
+                raise ValidationError(str(self.messages['choices'])
```
jaysonsantos: good idea, I will implement it

schematics/types/compound.py
```diff
@@ -128,7 +129,7 @@ def _convert(self, value, context):
             model_class = self.model_class
         else:
             raise ConversionError(
-                "Input must be a mapping or '%s' instance" % self.model_class.__name__)
+                _("Input must be a mapping or '%s' instance", lazy=False) % self.model_class.__name__)
```
jaysonsantos: if we implement `__mod__` and `format` and we use it we should be ok

schematics/types/base.py
```diff
@@ -394,24 +393,24 @@ def to_native(self, value, context=None):
...
             else:
                 return str(value)
-        raise ConversionError(str(self.messages['convert']).format(value))
+        raise ConversionError(self.messages['convert'].format(value))
 
     def validate_length(self, value, context=None):
         length = len(value)
         if self.max_length is not None and length > self.max_length:
-            raise ValidationError(self.messages['max_length'])
+            raise ValidationError(str(self.messages['max_length']))
```
lkraider: Are these casts required, since you are casting inside exceptions.FieldError ?

schematics/translator.py
```diff
@@ -1,3 +1,6 @@
+# coding: utf-8
+
+
 class LazyText(object):
```
lkraider: The import and class decorator are still missing.

schematics/types/base.py
jaysonsantos: I missed some, I will fix it now.

lkraider:
Perfect, thanks so much for the contribution!

PR #496, opened 2017-06-13
Add ability to import models as strings

jaysonsantos:
This would make it 'lazy' so you can import stuff and not have circular imports if required I can also add a post setup function which could validate this.


jaysonsantos:
I just rebased this.

jaysonsantos:
Hi, any news on this? I have rebased it again to solve conflicts.

PR #502, opened 2017-06-23
Use built-in __subclasses__() instead of registering

mjrk:


bintoro:
`C.__subclasses__()` only returns the immediate subclasses of `C`. The purpose of the registration procedure is to provide a list of *all* descendant classes without resorting to a recursive `__subclasses__()` call on the fly. I'm almost sure that's the only reason `__subclasses__()` wasn't used here in the first place.

So, if you're going to use `__subclasses__()`, it's got to be recursive.

mjrk:
@bintoro yes, https://github.com/schematics/schematics/pull/502/files#diff-bca9ce3cb5abd34125595fe78191cd1aR85 is recursive. (accidentally broken when solving merge conflicts.)

If I understood correctly, these subclasses are basically required for the polymodel functionality. IMHO it's okay then that they don't effect all the other models. We we don't want to calculate these every time, we can memoize them first access. On class level? Or instance level, just memoizing the candidates for the poly model in https://github.com/schematics/schematics/pull/502/commits/6e5718a64344be23a6f9d76d133aca181dc86444#diff-b313ceebdeffc1af782624306c84d9c1R427

What do you think?

bintoro:
Wouldn't memoization lead to the same memory issue that started this whole thing? The performance hit from doing a recursive lookup every time is probably negligible in realistic scenarios, so maybe we could just go that route for now.

mjrk:
True, we can only memoize on instance level to avoid any memory issues. But it would solve our memory issue because we don't use poly models for the dynamic ones and the memoization would only take place there.

Still I'd go for no memoizaion at all - if somebody needs it, it would be easy to add by overriding the _get_candidates method.

PR #503, opened 2017-06-24
Use built-in __subclasses__() instead of registering

mjrk:


mjrk:
The error in travis is 

```
ERROR: InterpreterNotFound: python2.6
```

Can I fix this actually?

PR #449, opened 2016-09-17
Allow ModelOptions to have extra parameters

rmb938:
Closes #397 

ModelOptions can now have extra parameters. This only applies to params that start with '_'

Usage:

``` python
>>> class Foo(Model):
...     class Options:
...         _test = 123
... 
>>> Foo.Options._test
123
>>> Foo._options._test
123
```

Issue #397, opened 2016-03-03
ModelOptions should allow anykind of extra parameters


Guibod:
Got a `TypeError: __init__() got an unexpected keyword argument 'route'` error while trying to extend the Options class from my schematics model.

``` python
class Search(Model):
    uuid = UUIDType(required=True)
    revision = NumberType(required=True)
    created_at = DateTimeType(default=datetime.datetime.now)
    updated_at = DateTimeType(default=datetime.datetime.now)

    class Options:
        route = 'my_flask_route_name'
        namespace = 'searches'
        roles = {}
```

I guess that I can start my own implementation of `ModelOptions` and alter my `__optionsclass__` but that sucks.
Could you please allow an `extra` field that allow us to put all kind of stuff in the model Options ?

bintoro:
I totally agree.

The options code currently seems to special-case members starting with "_", but it doesn't actually set them on the hidden `_options` object:

```python
>>> class Foo(Model):
...     class Options:
...         _test = 123
... 
>>> Foo.Options._test
123
>>> Foo._options._test
AttributeError
```

I believe the original intent is just to skip `__special__` attributes. But we could build on this and allow custom members beginning with an underscore and only skip those that are *actually* special.

rmb938:
See pull #449 for the implementation of @bintoro 's comment.

PR #506, opened 2017-07-25
Feature/allow custom options

ekampf:
I think Model's Options should allow defining custom options without having to derive a new class from `SchemaOptions` and specifying `__optionsclass__` which is an unnecessary hassle...

The docs actually provide the following example ([see here](http://schematics.readthedocs.io/en/latest/usage/models.html#model-configuration)):
```
class Whatever(Model):
    ...
    class Options:
        option = value
```

Why not support it?  it's a lot nicer and straightforward than:
```
class MyOptions(SchemaOption):
  def __init(foo=None, **kwargs):
    super(MyOptions, self).__init__(**kwargs)
    self.foo = foo

class Whatever(Model):
  __optionsclass__ = MyOptions
  class Options:
    foo = "Bar"
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ docs/_build
 .coverage
 *.sqlite3
 /.idea
+/.venv
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ docs/_build
 /.cache
 .coverage
 *.sqlite3
+/.idea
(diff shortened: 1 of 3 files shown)


ekampf:
This is also useful in case we have several different modules adding functionality to `Model` where each expect its own set of optional configurations via the options structure. 
They can't each define their own`SchemaOptions` class to use because there can be only one...

Thats why `SchemaOptions` should be allowed to work as a generic data bag...

lkraider:
Thanks for the PR + test!

This is related to issue #397 and PR #449 . Since this have come up a few times now, I'll add it in. To avoid naming conflicts, I'll go with the solution proposed in the previous PR.

PR #508, opened 2017-08-10
Make role send more parameters and allow serialize send a callable role

jaysonsantos:
This helps when you have cases where you don't know at the definition
of the class which fields should be shown, for example an ACL which
requires a logged in user which could be sent as callable role to
serialize

### DIFF
--- a/schematics/role.py
+++ b/schematics/role.py
@@ -1,3 +1,5 @@
+import warnings
+
 from .compat import str_compat, repr_compat
 
 import collections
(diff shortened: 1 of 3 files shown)


lkraider:
Hi, thanks for the use case!

I am not too keen on changing the interface as proposed here, nor the double check on the transform loop, since that is the most called function in schematics.

As a solution, I have added support for passing a callable role, and I propose to use it like this:
```python
def dynamic_output(user_id, model):
    acl_fields = {
        'user_id_1': ['name'],
        'user_id_2': ['name', 'password'],
    }
    acl_role = whitelist(*acl_fields[user_id])
    return model.serialize(role=acl_role)  # change output depending on current user

```

Let me know if that's good enough and I'll close this PR.

lkraider:
Released in 2.1 !

Issue #510, opened 2017-08-14
Models can no longer be pickled


alexhayes:
It appears that somewhere between version `2.0.0a1` and `2.0.1` models can no longer be pickled.

```python
import pickle
from schematics.models import Model
from schematics.types import IntType

class Foo(Model):
    bar = IntType()

f = Foo(dict(bar=1))

pickle.dumps(f)
```

Results in: `TypeError: can't pickle mappingproxy objects`

Note that I attempted to do a `git bisect start HEAD a8249f1 --` on master but there are many commits that are essentially broken thus bisect doesn't appear possible. Even if I modify the above script to return `125` if we can't determine the exit code (ie.. for issues like `cannot import name 'ComputedType'` and `No module named 'schematics.inspection'`) essentially git bisect gets to a point where there are only skipped commits left to test (ie.. dead end).

With a bit more digging I could ascertain that it was definitely broken by 65f3d95 and was good in b90990a so somewhere between those two it was broken.

This is perhaps related to #474.

lkraider:
Thanks, I'll open a review for the pickling issues. It requires tests to make it a reliable feature.

alexhayes:
@lkraider I can open a PR with tests if you like.

lkraider:
All help is welcome! :)

alexhayes:
Just to be sure `MappingProxyType` is not supposed to support pickle I've raised http://bugs.python.org/msg300284

rooterkyberian:
For me this have caused problems during with deepcopying, so my workaround using [schematics 2.0.1](https://github.com/schematics/schematics/releases/tag/v2.0.1) is```py
def __deepcopy__(self, memo):
    copied = ModelDict(
        *map(lambda o: copy.deepcopy(o, memo),
        [self._unsafe, self._converted, self._ModelDict__valid]),
    )
    return copied

ModelDict.__deepcopy__ = __deepcopy__
```

it would be nice to get rid of this ugliness ASAP

**Comment by froxCZ** [comment]
I fixed it by implementing own Schematics base class and overriding pickle methods.
```
class PickableModel(schematics.Model):
    def __getstate__(self):
        # For pickle and depickle
        return self.to_native()

    def __setstate__(self, kwargs: Dict[str, any]):
        self.__init__(kwargs)
```

ses4j:
Problem persists into `2.1.0`

teck968:
May I request an update on the direction for this issue? 
[...]

PR #511, opened 2017-08-14
WIP - Refs #510 - Models can no longer be pickled

alexhayes:
See #510

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-
+import pickle
 import pytest
 
 from schematics.common import PY2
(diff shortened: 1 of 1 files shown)


alexhayes:
Just to be sure that `MappingProxyType` is not supposed to support pickle I've raised this http://bugs.python.org/msg300284

alexhayes:
Just to be sure `MappingProxyType` is not supposed to support pickle I've raised http://bugs.python.org/msg300284

lkraider:
I am thinking of instead of using different dicts, actually marking the values with metadata to know if they are valid or not. This would simplify the structure back to a single container which could expose iterators for data in each state.

alexhayes:
@lkraider given what you've said above will that be something that will be happening in the short term or is it more a mid-long term goal?

I'm just wondering if I should expect that this PR will be merged (in which case I might use my fork for the time being knowing that it will be merged) or if I should hold off upgrading from `2.0.0a1` altogether for the time being.

lkraider:
I want to fix this by removing the `MappingProxyType` yes, but I would recommend if you require pickling stick with what works for you right now. I'm updating the internals using this PR + tests, so this PR might not merge as-is afterwards.

Issue #513, opened 2017-08-31
Incorrect documentation RE: BaseType's serialize_when_none arg


tommyzli:
The [docstring](https://github.com/schematics/schematics/blob/v2.0.0.dev2/schematics/types/base.py#L144-L146) for the `BaseType`  indicates that the default value for the `serialize_when_none` argument should be `True`, but in the code it is actually [`None`](https://github.com/schematics/schematics/blob/v2.0.0.dev2/schematics/types/base.py#L168). The [documentation](https://schematics.readthedocs.io/en/latest/api/types.html?highlight=serialize_when_none#schematics.types.base.BaseType) is also inconsistent; the class definition accurately indicates that the default is `None`, but the expanded parameter definitions below still say it's default is `True`.

lkraider:
Thanks for the report, the documentation needs real love. All help is welcome.

tommyzli:
I'd be glad to contribute, if you can point me to where the documentation is generated from - the files in the `docs/` folder don't seem to match the readthedocs site.

lkraider:
You can edit the ones from the docs folder.

The site is pulling from the old development branch which has been removed.

@bintoro can you edit the settings there to build the docs from master, or add me there so I can do it?

PR #514, opened 2017-09-04
Fix BaseType's incorrect docstring

tommyzli:
This addresses https://github.com/schematics/schematics/issues/513.


lkraider:
Thanks!

Issue #515, opened 2017-09-18
Unexpected behaviour in DateType's to_native conversion


tommyzli:
If you pass a `datetime` object into a `DateType` field, the native conversion will not turn the datetime into a date.

To reproduce:
```python
from datetime import datetime
from schematics.types.base import DateType

date_type = DateType()
today = datetime.today()

print(date_type.to_native(today))
>>> datetime.datetime(2017, 9, 18, 15, 40, 55, 10313)
```

This is caused by [this](https://github.com/schematics/schematics/blob/master/schematics/types/base.py#L680) line, as  `datetime` is a subclass of `date`, and will pass that check. 

In my opinion, `DateType`'s to_native function should only return `date` objects, not `datetime`. If you folks find this agreeable, I can open a pr with a fix.

lkraider:
Indeed, `datetime` inherits from `date`, so the conversion passes those instances through. I agree it should cast datetime to date, adding something like this should do it:

```python
if isinstance(value, datetime.datetime):
     value = value.date()
```

If you create a PR I can merge it.

tommyzli:
@lkraider I've created a pull request with the fix: https://github.com/schematics/schematics/pull/516

lkraider:
Thanks!

PR #516, opened 2017-09-18
Addresses #515 - DateType's to_native now converts datetimes into dates

tommyzli:


lkraider:
Thanks for the patch!

PR #518, opened 2017-09-21
Net types documentation

rooterkyberian:
Very minimal, but at least it will be included in autogeneration from now on.

I have also messed with address types mocks while at it (IPv6 was bugged).


lkraider:
🏆

PR #520, opened 2017-09-22
support ListType(..., min_size=0) mocking, resolves #464

### LINKED ISSUE #464: ListType mock objects: min values for 'min_size' and 'max_size'
It's not possible to generate empty mock list using `ListType`, because it uses bool-like checks:

```python

min_size = self.min_size or 1
max_size = self.max_size or 1
```

Default value for these attributes is `None`, so it's better to compare values with `None`:

```python

min_size = self.min_size if self.min_size is not None else 1
max_size = self.max_size if self.max_size is not None else 1
```

Thanks

rooterkyberian:


lkraider:
Thanks!

Issue #522, opened 2017-09-29
PEP 484 type-hints


chadrik:
Hi,
Do you all have any interest in adding type hints to schematics?  I find adding them sort of meditative so I'd be glad to work on it for this project.  If you're interested, you have two choices for how to go about it:

## option 1: stubs
create .pyi stub files that describe the types of arguments of all public functions in the library.  these reside alongside their .py partners and are used by apps like PyCharm and mypy to give errors and completions when `schematics` is used within other projects.

### pros
- does not impact existing code at all
- does not require adding `typing` module as a requirement

### cons
- stubs have to be manually updated any time methods are added, removed, renamed, or the signature changes

## option 2: annotations
serves the same purpose as .pyi files, but the annotations live in the code instead of separate files.

### pros
- only one set of files to maintain
- good starting point for adding complete static type checking to schematics.  stubs are only useful for projects that *consume* `schematics`.  annotations used within `schematics` code could be used to statically type check `schematics` itself as part of the CI process.

### cons
- [comment-based type annotations](http://mypy.readthedocs.io/en/latest/python2.html) are fairly ugly.  you have to go to 3.5+ to get syntactical support.  I'm working on tools to convert comment-based type annotations to 3.5-compatible code, so when the project does eventually go 3.x-only you won't have to refactor it all by hand.
- requires adding `typing` module as a requirement for python 2.x
- may require inheriting from [`typing.Generic`](http://mypy.readthedocs.io/en/latest/generics.html#defining-generic-classes) in a few cases instead of `object`.

lkraider:
Hi @chadrik , this is definitely something that would be good to have. I've added type annotations in https://github.com/schematics/schematics/blob/master/schematics/iteration.py , I think this going way should work well for other parts too. Let me know what you think.

rileyjohngibbs:
Are there any designs on continuing with/restarting this project? It looks like the one PR connected to it never got completed.

PR #523, opened 2017-09-29
PEP 484 type checking

chadrik:
Related to: #522

I understand the desire to avoid adding `typing` as a requirement, but doing so does open up some powerful options that are not otherwise available:

- `Generic` classes
- `NamedTuple`
- `@overloads`

I demonstrated the use of some of these in this commit.  Let me know what you think.

### DIFF
--- a/schematics/iteration.py
+++ b/schematics/iteration.py
@@ -3,21 +3,22 @@ from .compat import iteritems
 from .undefined import Undefined
 from collections import namedtuple
 
-try:
-    # optional type checking
-    import typing
-    if typing.TYPE_CHECKING:
-        from typing import Mapping, Tuple, Callable, Optional, Any, Iterable
-        from .schema import Schema
-except ImportError:
-    pass
+from typing import *
 
-Atom = namedtuple('Atom', ('name', 'field', 'value'))
+if TYPE_CHECKING:
+    from .schema import Schema, Field
+
+
+Atom = NamedTuple('Atom',
+                  (('name', Optional[str]),
+                   ('field', Optional['Field']),
+                   ('value', Any)))
+
 Atom.__new__.__defaults__ = (None,) * len(Atom._fields)
 
 
 def atoms(schema, mapping, keys=Atom._fields, filter=None):
-    # type: (Schema, Mapping, Tuple[str, str, str], Optional[Callable[[Atom], bool]]) -> Iterable[Atom]
+    # type: (Schema, Mapping, Tuple[str, str, str], Optional[Callable[[Atom], bool]]) -> Iterator[Atom]
     """
     Iterator for the atomic components of a model definition and relevant
     data that creates a 3-tuple of the field's name, its type instance and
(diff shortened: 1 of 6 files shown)


chadrik:
Hi, it would be great to have some feedback on this.  Namely, whether it is acceptable to add `typing` as a requirement for python 2.x.  Personally, I think we'll be very restricted in what we can accomplish without it.

lkraider:
I agree with you, adding `typing` has enough benefits to merit the burden of bringing an extra dependency. Let's move forward with enabling it!

chadrik:
@lkraider glad to hear it. 

Question:  What is the state of the deprecated mixins?  I see in `ModelType._export()` that a `Model` class is passed to `export_loop()`, but `export_loop` and `import_loop` behave as if they were passed a `Schema` instance:  they on `ModelCompatibilityMixin` being injected into `Model`  in order to make a `Model` class appear as if it is a `Schema` instance.  

Should `export_loop/import_loop` expect `Schema`, `Type[Model]` or `Union[Schema, Type[Model]]`?

chadrik:
Here's the code I'm referring to: https://github.com/schematics/schematics/blob/master/schematics/types/compound.py#L154

You'll notice that `ModelType._convert` goes through the model class which ultimately ends up passing in its `Schema` object:

```python
model_class.convert(value, context=context)
```

... whereas `ModelType._export` calls directly into `export_loop` with the `Model` class instead of the `Schema`:

```python
export_loop(model_class, value, context=context)
```

Should this be `model_class.export()`?

lkraider:
I have not yet gone through the compound types. My idea was to phase out the compatibility mixins at the same time move the Model fields into state-machines that keep their own info of whether they are already validated or not.  The plan is to free us from having to pass the model around everywhere. But that is still only in the idea phase.

chadrik:
Hi, I had some time to hack away at this some more.   It's going to be a long process, so we should try to break it up into smaller tasks that can be merged as we go.  The first of these tasks that I've encountered, which I could use someone else's help on, is converting use of `__all__ = module_exports(__name__)` into lists of static strings.  I know, I know, I'm too lazy to maintain these lists manually, too, but if you use wild imports and `__all__`, then mypy must be able to statically inspect the names in `__all__`.  Shall I make a separate issue for this?

lkraider:
Yes let's do it. I am not a fan of the star import myself, but I think we should curate the exports.

chadrik:
Cool, any chance you can help me out with that?  Maybe the simplest thing to do is hack `module_exports` to pretty-print the list that it finds, then walk through and import all the modules.

lkraider:
Sure thing, I plan on working this weekend on the release, and I'll also work on the exports then.

lkraider:
I've made the changes to the master branch, let me know if it works this way.

ses4j:
Did this die on the vine?

HJEGeorge:
Bump - would be very keen on having schematics typing for both schematics models and primitives (some form of typed_dict).

PR #537, opened 2018-01-19
Simplify PolyModelType.find_model for readability

kstrauser:
This rearranges existing logic to be more Python and readable. In particular, it removes a level of nested if's and uses multiple returns to avoid tracing a flag variable through the function.

My coworker asked me how this worked, and it took uncomfortably long for me to be able to tell him. This might shorten that awkward silence for the next person. :-)

### DIFF
--- a/schematics/types/compound.py
+++ b/schematics/types/compound.py
@@ -392,31 +392,35 @@ class PolyModelType(CompoundType):
     def find_model(self, data):
         """Finds the intended type by consulting potential classes or `claim_function`."""
 
-        chosen_class = None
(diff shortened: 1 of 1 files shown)


kstrauser:
I think this fixes #537, too.

kstrauser:
@lkraider This is a pretty simple change that doesn't affect any client code. Think we could sneak this into the upcoming release?

lkraider:
Thanks! I am organizing a 2.1 release, this will be in it!

kstrauser:
Awesome, thanks!

Issue #539, opened 2018-02-05
Create a Timedelta/Duration type


gabisurita:
Following what is implemented for ` DateTimeType` and `DateType`, I think it would be nice to have a `TimeDeltaType` for `datetime.timedelta` objects.

lkraider:
Indeed, a PR is welcome!

gabisurita:
Nice! I'll try to sketch something up later today.

lkraider:
Released in 2.1 !

PR #540, opened 2018-02-15
Implement TimedeltaType

gabisurita:
Implements https://github.com/schematics/schematics/issues/539

Implement a TimeDeltaType matching native`datetime.timedelta` type.

Also simplifies `TimeStampType.to_native` call.

### DIFF
--- a/schematics/types/base.py
+++ b/schematics/types/base.py
@@ -966,8 +966,7 @@ class TimestampType(DateTimeType):
         else:
             value = value.astimezone(self.UTC)
         delta = value - self.EPOCH
-        ts = (delta.days * 24 * 3600) + delta.seconds + delta.microseconds / 1E6
-        return ts
+        return delta.total_seconds()
 
 
 class TimedeltaType(BaseType):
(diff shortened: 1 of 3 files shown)


schematics/types/base.py
```diff
@@ -965,8 +966,39 @@ def to_primitive(self, value, context=None):
         else:
             value = value.astimezone(self.UTC)
         delta = value - self.EPOCH
-        ts = (delta.days * 24 * 3600) + delta.seconds + delta.microseconds / 1E6
-        return ts
+        return delta.total_seconds()
```
gabisurita: This is not related to the fix itself, but is a way more explicit way to get this value.

schematics/types/base.py
```diff
@@ -965,8 +966,39 @@ def to_primitive(self, value, context=None):
...
+        if isinstance(value, datetime.timedelta):
+            return value
+
+        try:
+            return datetime.timedelta(seconds=float(value))
+        except (ValueError, TypeError):
+            raise ConversionError(self.messages['convert'].format(value))
+
+    def to_primitive(self, value, context=None):
+        return value.total_seconds()
```
lkraider: Must return the value according to the selected init option here. Can divide by `timedelta(**{self.precision: 1})` to get the same scale as input.

tests/test_types.py
```diff
@@ -132,6 +132,39 @@ def test_datetime_accepts_datetime():
     assert dt_type.to_primitive(dt) == output
 
 
+def test_timedelta():
+    type_ = TimedeltaType()
```
lkraider: I don't think assigning to an intermediary `type_` variable is necessary in these tests -- I see some other tests use an intermediary, but usually when the type is slow to init or many versions/init options are used, so as not to repeat them.

Issue #527, opened 2017-11-07
Schema Doc Gen?


ben174:
Are there any packages that will self-document all models defined in a project? Possibly create HTML output or something of the sort?

lkraider:
That would be pretty cool. Maybe as a plugin for Sphinx?

I found this example for Django models: https://gist.github.com/abulka/48b54ea4cbc7eb014308

ben174:
Awesome! Might take this on. Will keep you posted in here if I end up moving forward on this.

PR #543, opened 2018-03-03
Add help text generation capability to Models

MartinHowarth:
This addresses [#527](https://github.com/schematics/schematics/issues/527)

Adds three new methods to the Model as below, with example output
- `get_helptext` - Generate friendly help text (which I see as being useful for a CLI tool help output):

```
This model describes a person.
  name (Name)
    Example: Joe Stummer
    Full name.
  age (Age)
    Example: 24
    Age, in years.
```

- `get_example_usage` - Generate sample python code to use the model:

```
Person({
    'name': Joe Stummer,
    'age': 24,
})
```

- `get_api_docstring` - Generate sphinx api-doc compatible docstring:

```
"""
This model describes a person.


Example:

    Person({
        'name': Joe Stummer,
        'age': 24,
    })


:param str name: Full name.
:param int age: Age, in years.
"""
```

Tempted to refactor it a bit before merging, but thought I'd ask for some feedback on what I've got working already - let me know what you think.

Also made the standard `python setup.py test` command work (thought it was a bit odd that this was missing, but feel free to delete those changes from this MR, or ask me to)

### DIFF
--- a/schematics/__init__.py
+++ b/schematics/__init__.py
@@ -6,7 +6,7 @@ __version__ = '2.0.1'
 from . import deprecated
 deprecated.patch_all()
 
-from . import types
+from . import types, extensions
 from .models import Model, ModelMeta
 
 types.compound.Model = Model
--- a/schematics/extensions/model_help_text.py
+++ b/schematics/extensions/model_help_text.py
@@ -1,0 +1,120 @@
+from textwrap import fill
+
+from ..iteration import atoms
+
+
+def help_text_metadata(label=None, description=None, example=None):
+    """
+    Standard interface to help specify the required metadata fields for helptext to
+    work correctly for a model.
+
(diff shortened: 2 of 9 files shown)


lkraider:
Hi thanks for the patch!

Just did a quick look over it, what do you think of providing it as a mixin for the Model class? I've been meaning to do that for the mock functionality too. I suppose we might as well create an "extensions" folder to place these.  Might be good to have a test file per mixin as well.

Also, seems Travis is not happy with something on Python 2.7, might need to check that.

MartinHowarth:
Thanks for getting back to me :)

I liked your suggestion of making this a mixin instead, so I went ahead and did that.

What do you think of it now?

MartinHowarth:
Hi, what's the status of this? 😄

PR #550, opened 2018-03-19
Set default when value is None

### LINKED ISSUE #548: Use default value when value is None
Using the default keyword is convenient.
```python
class Person(Model):
    name = StringType(default='Guido')

p = Person() 
print(p.name) 
>>> 'Guido'
```
However, if the value gets set to None, the default isn't triggered.

```python
p = Person({'name': None}) 
print(p.name)
>>>'None'
```
This makes it hard to abstract away the instantiation of a model.
Ideally the default value should always be set when the value is None.

Use case:
When getting a request from the web that you'd like to parse, the request will look something like this:
`{"name": None}`.  If you dump this into your Schematics Model, the default will never get set.  

In this case, subclassing the type and overriding the `convert`, `to_native`, or `to_primitive` methods doesn't work, because they never get called when a `None` value is passed in.

leahein:
Resolves #548 

## Added: 
This will set a field to its default value when the field is passed into the instance with a value of `None`.
### For example:
```python
class M(Model):
    field = StringValue(default=0)
    field_two = StringValue()
```

__Past behavior__:
```python
m = M({'field': None, 'field_two': None})
m.field
>>> None
m.field_two
>>> None
```

__PR Update__: 
```python
m = M({'field': None, 'field_two': None})
m.field
>>> 0
m.field_two
>>> None
```
### Use Case:
This allows web requests to be parsed easily by passing a dictionary request into the model, and ensuring that the default values will be set.

### DIFF
--- a/schematics/transforms.py
+++ b/schematics/transforms.py
@@ -134,6 +134,10 @@ def import_loop(schema, mutable, raw_data=None, field_converter=None, trusted_da
         if value is Undefined and context.init_values:
             value = None
 
+
+        if context.apply_defaults and value is None and field.default is not Undefined:
+            value = field.default
+
         if got_data:
             if field.is_compound:
                 if context.trusted_data and context.recursive:
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -116,6 +116,17 @@ def test_defaults():
     assert m._data == {'d0': 0, 'dN': None}
 
 
+    m = M({'d0': None}, apply_defaults=True)
+    assert m._data == {'d0': 0, 'dN': None}
+    m.validate()
+    assert m._data == {'d0': 0, 'dN': None}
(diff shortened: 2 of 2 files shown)


bintoro:
Schematics is designed to apply defaults to those fields that are not present in the input data. If `None` is equated with "not there", it would then become impossible to explicitly initialize a field to `None`. The field would always end up with the default value instead, with no way to override it.

leahein:
Thank you for following up.

Yes, I agree that not allowing an override of `None` for the defaults may not be the behavior that is always wanted.  
In a case where there's a schema that gets passed in with `None` values and you would like for the defaults to be set, would you know if there's a good way of handling that? 
If not, is there a way to leverage the library in its current state to achieve that behavior?

Issue #566, opened 2018-08-13
UnionType with ModelType fails during Model.to_primitive() with AttributeError: 'Context' object has no attribute 'convert'


acormier128:
Attempting to use ModelType as a type within a UnionType seems to work except when trying to convert the model back into a dict via to_primitive():

The test: 

```
import pytest
from schematics import models, types, undefined

class SimpleModel(models.Model):
    simple_field = types.StringType()

class MyModel(models.Model):
    my_field = types.UnionType((types.StringType, types.ModelType), model_spec=SimpleModel)

def test_union_type():
    my_inst = MyModel(dict(my_field={'simple_field': 'simple_value'}))
    my_inst.to_primitive()
```

The error:
```
============================================================== FAILURES ===============================================================
___________________________________________________________ test_union_type ___________________________________________________________

    def test_union_type():
        my_inst = MyModel(dict(my_field={'simple_field': 'simple_value'}))
>       my_inst.to_primitive()

repro.py:12:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../../.venv_terrasphere/lib/python3.6/site-packages/schematics/models.py:309: in to_primitive
    return to_primitive(self._schema, self, role=role, app_data=app_data, **kwargs)
../../../../.venv_terrasphere/lib/python3.6/site-packages/schematics/transforms.py:436: in to_primitive
    return export_loop(cls, instance_or_dict, to_primitive_converter, **kwargs)
../../../../.venv_terrasphere/lib/python3.6/site-packages/schematics/transforms.py:277: in export_loop
    value = _field_converter(field, value, context)
../../../../.venv_terrasphere/lib/python3.6/site-packages/schematics/transforms.py:355: in __call__
    return self.func(*args)
../../../../.venv_terrasphere/lib/python3.6/site-packages/schematics/transforms.py:370: in to_primitive_converter
    return field.export(value, PRIMITIVE, context)
../../../../.venv_terrasphere/lib/python3.6/site-packages/schematics/types/base.py:286: in export
    return self.export_mapping[format](value, context)
../../../../.venv_terrasphere/lib/python3.6/site-packages/schematics/types/union.py:96: in to_primitive
[...]

Issue #556, opened 2018-04-27
Using multiple Model as alterantive?


cnicodeme:
Hi!

I'm trying to find a way to allow various Model as altenatives to handle and validate a field. I thought about UnionType or PolyModelType but I can't make them work.

Here's a sample:

```
class AlternativeAForm(Model):
    author         = StringType(required=False)

class AlternativeBForm(Model):
    user         = StringType(required=False)

class AlternativeCForm(Model):
    owner         = StringType(required=False)

class OriginalForm(Model):
    email  = StringType(required=True)
    name  = UnionType([ModelType(AlternativeAForm), ModelType(AlternativeBForm), ModelType(AlternativeCForm)])
```

This would accept JSON post like
```{"email": "...", "name": {"author": "..."}}```
or
```{"email": "...", "name": {"user": "..."}}```
or
```{"email": "...", "name": {"owner": "..."}}```

(My usage is more advanced than this example of course ;) ).

I haven't found where I could ask for help so I'm asking here, hope that's alright!

Best,

**Comment by naphta** [comment]
@cnicodeme Did you ever figure out a workaround to this?

**Comment by naphta** [comment]
@cnicodeme I think I've figured out the issue with using multiple model types; seems to be a bug in the UnionType code.

```
< from ..exceptions import ConversionError
---
> from ..exceptions import ConversionError, DataError
40d39
<
51,52c50,51
<             self._types[type_.__class__] = type_
<             self.typenames = tuple((cls.__name__ for cls in self._types))
---
>             self._types[repr(type_)] = type_
>             self.typenames = tuple((cls for cls in self._types))
60c59
<             except ConversionError:
---
>             except (ConversionError, DataError):
72c71
<             except ConversionError:
---
>             except (ConversionError, DataError):
```

that's the diff of what I'm currently testing; basically the ModelType wasn't raising ConversionError, and when the types were being set on `__init__` they were being added to a dictionary with `type_.__class__` which always resolved to `<class 'schematics.types.compound.ModelType'>`, I switched it to repr() as a workaround as that keeps the models name in it. Once I've got something working I'll open up a pull request though.

PR #582, opened 2019-01-16
fix export for UnionType that contains ModelType

azban:
fixes https://github.com/schematics/schematics/issues/566

i'm not sure if this is the best place to do this. it seems like maybe `export_mapping` could contain this information, but i'm not sure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ docs/_build
 /.cache
 .coverage
 *.sqlite3
+.idea/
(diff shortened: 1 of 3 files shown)


azban:
also fixes https://github.com/schematics/schematics/issues/556

this doesn't seem like the most elegant solution, but it seems best given what's exposed on `BaseType`. let me know if it's an issue to combine these two changes.

syndicut:
@lkraider are there any plans for this to be merged?

Issue #571, opened 2018-10-12
Deprecation warnings when using Pytest


ZatkovaV:
I use schematics to define models for requests and responses in simple Flask API. 
However, when running tests with Pytest, I encounter many ```SchematicsDeprecationWarning``` warnings. 

The warnings are as following:
```
...lib/python3.7/site-packages/schematics/transforms.py:103: SchematicsDeprecationWarning: Call to deprecated function _valid_input_keys.
  all_fields = schema._valid_input_keys
...lib/python3.7/site-packages/schematics/validate.py:91: SchematicsDeprecationWarning: Call to deprecated function _validator_functions.
  atom.name in schema._validator_functions
...lib/python3.7/site-packages/schematics/transforms.py:103: SchematicsDeprecationWarning: Call to deprecated function _valid_input_keys.
  all_fields = schema._valid_input_keys
...lib/python3.7/site-packages/schematics/transforms.py:249: SchematicsDeprecationWarning: Call to deprecated function _options.
  if schema._options.export_order:
...lib/python3.7/site-packages/schematics/transforms.py:255: SchematicsDeprecationWarning: Call to deprecated function _options.
  schema._options.roles.get(context.role))
...lib/python3.7/site-packages/schematics/transforms.py:261: SchematicsDeprecationWarning: Call to deprecated function _options.
  filter_func = schema._options.roles.get("default")
...lib/python3.7/site-packages/schematics/deprecated.py:65: SchematicsDeprecationWarning: Call to deprecated function _options.
  return super(class_property, self).__get__(type, type)
...lib/python3.7/site-packages/schematics/deprecated.py:65: SchematicsDeprecationWarning: Call to deprecated function _options.
  return super(class_property, self).__get__(type, type)
...lib/python3.7/site-packages/schematics/deprecated.py:65: SchematicsDeprecationWarning: Call to deprecated function _options.
  return super(class_property, self).__get__(type, type)
```

Is there something I can do in order to solve this issue? My schematics version is 2.1.0. Thanks!

rmmcnulty9:
I needed to set this env var to get around this. But this seems like a bad solution: `PYTHONWARNINGS=0`

jesuslosada:
@rmmcnulty9 @tomasbedrich Note that #576 only fixes the deprecation warnings related to the standard library. There are additional warnings generated by Schematics because some features are deprecated but are still being used (?). You can at least ignore those errors passing this parameter to pytest:
[...]

PR #595, opened 2019-06-16
Removed usages of deprecated fields

kamikaz1k:
(preface) Hi o/ not sure if you're looking for PRs in this category of work but I saw https://github.com/schematics/schematics/issues/571 and it was annoying me in my builds, so I thought I would try to solve it instead of piling on. That being said, I did a very naive refactor based on what I understood. Ya'll might have had more comprehensive plans about how you wanted to arrange things. But hopefully this is roughly in the ballpark. (/preface)

There were a number of deprecation errors being emitted from the package, so I have attempted to update all those references by essentially taking the wrapping code and moving it closer to the calling site. I did my best with the understanding of the interactions.

I have also added a `Makefile` that makes is easier to set up a dev environment and get tests running. I realize I am imposing this, so let me know if you would rather not have this.

#### Makefile output:
```bash
$ make
=======================================================
=============== schematics dev commands ===============
=======================================================
env 		setup virtualenv
tests 		setup virtualenv and run tests in it

$ make env
test -d venv || virtualenv venv || @echo "please install virtualenv"
venv/bin/pip install -r test-requirements.txt
... pip do the installs ...
Successfully installed schematics

$ make test
... does `make env`
coverage pytest ... # the command from the README
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,21 @@
+.PHONY: help env test
+
+VENV=venv
+PIP=$(VENV)/bin/pip
+COVERAGE=$(VENV)/bin/coverage
+
+help:
+	@echo "======================================================="
+	@echo "=============== schematics dev commands ==============="
+	@echo "======================================================="
+	@echo "env \t\tsetup virtualenv"
+	@echo "tests \t\tsetup virtualenv and run tests in it"
+
+env:
+	test -d $(VENV) || virtualenv $(VENV) || @echo "please install virtualenv"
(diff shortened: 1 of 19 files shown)


jimorie:
+1 from me to get this or a similar fix merged.

This greatly reduces the number of deprecation warning I see in my schematics project. (Still have one call to deprecated function fields, from deprecated.py line 65.)

dhallam:
Is there a maintainer who can review, merge and release a 2.1.1 if this is okay?

PR #619, opened 2021-04-03
All in on Python 3

kstrauser:
Hi gang! It's been a while!

Python 2 is dead and Python 3 is the present. This PR removes Python 2 compatibility and starts adopting Python 3 features. It:

- Updates tox to use currently supported versions of Python
- Removes `compat.py`
- Switches to Python 3-style classes (i.e. `class Foo:` instead of `class Foo(object):`)
- Removes all `__future__` imports (the future is here!)
- Removes `deprecated.py` (as long as we're breaking things...)
- Uses the new "bare" `super()` syntax
- Adds just a few pylint and isort cleanups (more to follow in a separate PR)
- Mentions that I've contributed a line or two from time to time 😉
- Whenever reasonable, switches to f-strings (but I wasn't a purist about it)

These are breaking changes, and I think this would have to go into a new release: Schematics 3. That's a happy semantic versioning coincidence, isn't it?

There are some other things I'd like to do before we'd consider cutting a new release, but this PR is big enough as it is so I'll put those in other PRs.


Mizaro:
Do you think this project can stay python2 compatible?

kstrauser:
I don't think we can. Python 2 is officially 100% dead and unsupported as of a year ago, and sticking with that means we can't use the new features available in more recent versions. I see literally no downside to dropping Python 2 support so that we can make Python 3 support better.

lkraider:
LGTM. Great work @kstrauser ! 

Moving to Python 3 is important to keep maintenance manageable, contributions are scarce as it is.

@Mizaro is there some existing PR you would request for a last Python2 release?

kstrauser:
@lkraider Thanks, and I'm glad you liked it! It could have gone another way: "thanks for dropping this giant ball of mud PR on me, un-asked for." 😀

As a practical matter, since this is a very backward-incompatible change, I could see a release process like this:

- Rename the current `master` branch to something like `v2` or `schematics2` or such. Then we could continue to easily merge PRs onto that if appropriate, say if someone fixed a Python 2-specific bug.
- I opened this PR to merge into a new `schematics3` branch, but maybe we rename that to `main`, `dev`, or similar so that if people were previously tracking `master` it would break their build instead of merging in a new likely incompatible version. 

Then the main branch of Schematics would be on the new Python 3 version, but the old version would still exist for people who need it.

What do you think? Please take the above in the spirit of "Kirk brainstorming ways to make this as easy as possible for everyone", not "Kirk trying to dictate how to run a project he's only active with every few years". 🙂

hugovk:
> As a practical matter, ...

I'd recommend keeping branches as is, and using tags for versions. That would be much cleaner. 


> * Then we could continue to easily merge PRs onto that if appropriate, say if someone fixed a Python 2-specific bug.

Are you considering still maintaining for Python 2? I'd suggest drawing a line and not bothering at this point, it's been dead 19 months and many projects have dropped it. And this PR is titled "All in on Python 3" :) 


> Then the main branch of Schematics would be on the new Python 3 version, but the old version would still exist for people who need it.

The old version will still exist anyway on PyPI and in the repo via tags. Make sure to add `python_requires=">=3.6",` to `setup.py`, that way pip will know to install the new version for 3.6+ and will use the previous version for anyone still on 2.7 etc.

kstrauser:
@hugovk You've convinced me.

yesyouken:
Are there any plans to merge this in?

PR #620, opened 2021-04-03
Use pyproject.toml

kstrauser:
Python packaging is moving toward using [pyproject.toml](https://www.python.org/dev/peps/pep-0631/) to replace setup.py, requirements.txt, and other related files. This PR makes that change by:

- Using [Poetry](https://python-poetry.org) to create a new pyproject.toml
- Moving `test-requirements.txt` into the same file
- Moving `MANIFEST.in` into the same file
- Updating tox to use poetry to build the test environments
- Replacing the old `mock` development dependency with the built-in `unittest.mock` while mucking around with all of it (one less dependency!)

This builds on #619 so it's currently based off that PR. Once 619 is merged this should be rebased off that branch.


Mizaro:
Can you leave this python2 compatible?

PR #625, opened 2021-08-18
Add metadata to setup.py

hugovk:
Two things:

* Add `long_description`:
  https://pypi.org/project/schematics/ says only "The author of this package has not provided a project description". Instead, it's common to put the README contents in the `long_description` so it shows up on PyPI. Modern versions of setuptools, wheel and twine are needed, more info: https://packaging.python.org/guides/making-a-pypi-friendly-readme/

* Add `python_requires`:
  This helps pip choose the relevant version of this library for the user's running Python version. When EOL Python versions are dropped, set this to `python_requires='>=3.6`.


hugovk:
Closing this year old PR, happy to re-open if any interest 👍

